PR #16, opened 2024-07-15
Move query categorization changes to plugin

deshsidd:
Original PR : https://github.com/opensearch-project/OpenSearch/pull/14528

Query categorization changes to increment counters for search query related metrics currently resides on the search path and occurs before the request.

Move these changes to the query insights plugin and make sure the incrementing of counters happens separately from the search path.

Another option is to keep query categorization changes as is. However, this will lead to additional overhead on the search path. Furthermore, we need to tie query latency, cpu, memory with the query categorization data which will only be possible if we increment the counters after the request is completed and the query latency and resource usage data resides inside the plugin.

To support the above and to prevent doing these counter increments on the search path, we need to move query categorization changes to the query insights plugin.

Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/14527
Addresses https://github.com/opensearch-project/OpenSearch/issues/11596

Check List
 Functionality includes testing.
 API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
 Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -91,28 +91,27 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
      * and query insights services.
      *
      * @param metricType {@link MetricType}
-     * @param enabled    boolean
+     * @param isCurrentMetricEnabled boolean
      */
-    public void setEnableTopQueries(final MetricType metricType, final boolean enabled) {
-        boolean isInsightsServiceDisabled = !queryInsightsService.isEnabled();
+    public void setEnableTopQueries(final MetricType metricType, final boolean isCurrentMetricEnabled) {
+        boolean isTopNFeatureDisabled = !queryInsightsService.isTopNFeatureEnabled();
+        this.queryInsightsService.enableCollection(metricType, isCurrentMetricEnabled);
 
-        if (!enabled) {
+        if (!isCurrentMetricEnabled) {
             // disable QueryInsightsListener only if all metrics collections are disabled now
             // and search query metrics is disabled.
-            if (isInsightsServiceDisabled) {
+            if (isTopNFeatureDisabled) {
                 super.setEnabled(false);
-                this.queryInsightsService.stop();
+                queryInsightsService.checkAndStopQueryInsights();
             }
         } else {
             super.setEnabled(true);
             // restart QueryInsightsListener only if none of metrics collections is enabled before and
             // search query metrics is disabled before.
-            if (isInsightsServiceDisabled) {
-                this.queryInsightsService.stop();
-                this.queryInsightsService.start();
+            if (isTopNFeatureDisabled) {
+                queryInsightsService.checkAndRestartQueryInsights();
             }
         }
-
     }
 
     @Override
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -94,18 +94,20 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
      * @param enabled    boolean
      */
     public void setEnableTopQueries(final MetricType metricType, final boolean enabled) {
-        boolean isAllMetricsDisabled = !queryInsightsService.isEnabled();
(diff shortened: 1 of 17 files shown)


src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -94,18 +94,20 @@ public QueryInsightsListener(final ClusterService clusterService, final QueryIns
      * @param enabled    boolean
      */
     public void setEnableTopQueries(final MetricType metricType, final boolean enabled) {
-        boolean isAllMetricsDisabled = !queryInsightsService.isEnabled();
-        this.queryInsightsService.enableCollection(metricType, enabled);
```
ansjcy, line 98: We still need to call `this.queryInsightsService.enableCollection` to enable collection for a metric, it cannot be removed.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -239,6 +269,30 @@ public void setExporter(final MetricType type, final Settings settings) {
         }
     }
 
+    /**
+     * Set search query metrics enabled to enable collection of search query categorization metrics
+     * @param searchQueryMetricsEnabled boolean flag
+     */
+    public void setSearchQueryMetricsEnabled(boolean searchQueryMetricsEnabled) {
+        this.searchQueryMetricsEnabled = searchQueryMetricsEnabled;
```
ansjcy, line 277: In the `setSearchQueryMetricsEnabled` function, you might also need to check if we want to stop / restart the whole query insights service. We should write a generic function for this, like:
```
public void checkAndStopQueryInsights(...) {

}

    
public void checkAndRestartQueryInsights(...) {

}

 public void setSearchQueryMetricsEnabled(boolean searchQueryMetricsEnabled) {
    if (searchQueryMetricsEnabled) {
        ...
        checkAndRestartQueryInsights(...)
    } else {
        ...
        checkAndStopQueryInsights(...)
    }
 }
 public void setEnableTopQueries(final MetricType metricType, final boolean enabled) {
        ...
        if (!enabled) {
            ...
            checkAndStopQueryInsights(...)
        } else {
           ...
           checkAndRestartQueryInsights(...)
        }
    }
```

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: Got remove by mistake, good catch

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
deshsidd: Done in the latest commit

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -91,26 +91,27 @@ public QueryInsightsListener(final ClusterService clusterService, final QueryIns
...
-            // disable QueryInsightsListener only if all metrics collections are disabled now.
-            if (!queryInsightsService.isEnabled()) {
+    public void setEnableTopQueries(final MetricType metricType, final boolean isCurrentMetricEnabled) {
+        boolean isTopNFeatureDisabled = !queryInsightsService.isTopNFeatureEnabled();
+        this.queryInsightsService.enableCollection(metricType, isCurrentMetricEnabled);
+
+        if (!isCurrentMetricEnabled) {
+            // disable QueryInsightsListener only if all metrics collections are disabled now
+            // and search query metrics is disabled.
+            if (isTopNFeatureDisabled) {
```
ansjcy, line 103: should this be `!queryInsightsService.isTopNFeatureEnabled()` ? 
`isTopNFeatureDisabled` is the flag we get before changing a flag in `this.queryInsightsService.enableCollection`. At this time the flag as already be changed.

deshsidd: Added some unit tests to make sure we test this logic and refactored the code to make it clearer

ansjcy (approved):
LGTM, thanks for refactoring the enable functions and adding the tests!

PR #44, opened 2024-07-23
Query shape for agg & sort

dzane17:
### Description
Add ability to generate query shape for `aggregation` and `sort` portions of search body.

### Testing
Large Search
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "order_date": {
              "gte": "2023-01-01",
              "lte": "2023-12-31"
            }
          }
        },
        {
          "term": {
            "status": "completed"
          }
        }
      ],
      "filter": [
        {
          "terms": {
            "category": ["electronics", "clothing"]
          }
        }
      ]
    }
  },
  "aggs": {
    "sales_per_month": {
      "date_histogram": {
        "field": "order_date",
        "calendar_interval": "month",
        "format": "yyyy-MM",
        "order": {
          "_key": "desc"
        }
      },
      "aggs": {
        "total_sales": {
          "sum": {
            "field": "total_amount"
          }
        },
        "avg_order_quantity": {
          "avg": {
            "field": "order_quantity"
          }
        },
        "max_order_amount": {
          "max": {
            "field": "total_amount"
          }
        }
      }
    },
    "avg_monthly_sales": {
      "avg_bucket": {
        "buckets_path": "sales_per_month>total_sales"
      }
    },
    "avg_max_order_amount": {
      "avg_bucket": {
        "buckets_path": "sales_per_month>max_order_amount"
      }
    }
  },
  "sort": [
    { "sales_per_month>key": { "order": "desc" } },
    { "total_sales": { "order": "desc" } },
    { "avg_order_quantity": { "order": "asc" } }
  ]
}'
```
```
[2024-07-30T15:43:31,782][INFO ][stdout                   ] [integTest-0] bool []
[2024-07-30T15:43:31,783][INFO ][stdout                   ] [integTest-0]   must:
[2024-07-30T15:43:31,784][INFO ][stdout                   ] [integTest-0]     range [order_date]
[2024-07-30T15:43:31,784][INFO ][stdout                   ] [integTest-0]     term [status]
[2024-07-30T15:43:31,784][INFO ][stdout                   ] [integTest-0]   filter:
[2024-07-30T15:43:31,784][INFO ][stdout                   ] [integTest-0]     terms [category]
[2024-07-30T15:43:31,785][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-07-30T15:43:31,785][INFO ][stdout                   ] [integTest-0]   date_histogram [order_date]
[2024-07-30T15:43:31,785][INFO ][stdout                   ] [integTest-0]     aggregation:
[2024-07-30T15:43:31,785][INFO ][stdout                   ] [integTest-0]       avg [order_quantity]
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]       max [total_amount]
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]       sum [total_amount]
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]   pipeline aggregation:
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]     avg_bucket
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]     avg_bucket
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0] sort:
[2024-07-30T15:43:31,786][INFO ][stdout                   ] [integTest-0]   asc [avg_order_quantity]
[2024-07-30T15:43:31,787][INFO ][stdout                   ] [integTest-0]   desc [sales_per_month>key]
[2024-07-30T15:43:31,787][INFO ][stdout                   ] [integTest-0]   desc [total_sales]
```
Aggregation
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "aggs": {
    "group_by_category_4": {
      "terms": {
        "field": "category.keyword4"
      }
    },
    "group_by_category_3": {
      "terms": {
        "field": "category.keyword3"
      }
    },
    "group_by_category_2": {
      "terms": {
        "field": "category.keyword2"
      }
    },
    "group_by_category_1": {
      "terms": {
        "field": "category.keyword1"
      }
    }
  },
  "size": 10
}'
```
```
[2024-07-30T15:44:11,820][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-07-30T15:44:11,822][INFO ][stdout                   ] [integTest-0]   terms [category.keyword1]
[2024-07-30T15:44:11,822][INFO ][stdout                   ] [integTest-0]   terms [category.keyword2]
[2024-07-30T15:44:11,822][INFO ][stdout                   ] [integTest-0]   terms [category.keyword3]
[2024-07-30T15:44:11,822][INFO ][stdout                   ] [integTest-0]   terms [category.keyword4]
```
Nested Aggregation
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "aggs": {
    "sales_by_category": {
      "terms": {
        "field": "category.keyword",
        "size": 10
      },
      "aggs": {
        "total_sales": {
          "sum": {
            "field": "total_amount"
          }
        },
        "top_products": {
          "nested": {
            "path": "items"
          },
          "aggs": {
            "product_names": {
              "terms": {
                "field": "items.product_name.keyword",
                "size": 5
              },
              "aggs": {
                "avg_price": {
                  "avg": {
                    "field": "items.price"
                  }
                }
              }
            }
          }
        }
      }
    }
  },
  "size": 10
}'
```
```
[2024-07-30T15:44:16,824][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]   terms [category.keyword]
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]     aggregation:
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]       nested []
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]         aggregation:
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]           terms [items.product_name.keyword]
[2024-07-30T15:44:16,825][INFO ][stdout                   ] [integTest-0]             aggregation:
[2024-07-30T15:44:16,826][INFO ][stdout                   ] [integTest-0]               avg [items.price]
[2024-07-30T15:44:16,826][INFO ][stdout                   ] [integTest-0]       sum [total_amount]
```
Pipeline Aggregation
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "size": 0,
  "aggs": {
    "monthly_sales": {
      "date_histogram": {
        "field": "timestamp",
        "calendar_interval": "1M"
      },
      "aggs": {
        "avg_price": {
          "avg": {
            "field": "price"
          }
        },
        "cumulative_sum_price": {
          "cumulative_sum": {
            "buckets_path": "avg_price"
          }
        },
        "moving_avg_price": {
          "moving_avg": {
            "buckets_path": "cumulative_sum_price",
            "window": 3,
            "model": "simple"
          }
        }
      }
    }
  }
}'
```
```
[2024-07-30T15:44:26,831][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-07-30T15:44:26,831][INFO ][stdout                   ] [integTest-0]   date_histogram [timestamp]
[2024-07-30T15:44:26,832][INFO ][stdout                   ] [integTest-0]     aggregation:
[2024-07-30T15:44:26,832][INFO ][stdout                   ] [integTest-0]       avg [price]
[2024-07-30T15:44:26,832][INFO ][stdout                   ] [integTest-0]       pipeline aggregation:
[2024-07-30T15:44:26,832][INFO ][stdout                   ] [integTest-0]         cumulative_sum
[2024-07-30T15:44:26,832][INFO ][stdout                   ] [integTest-0]         moving_avg
```

Sort
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "query": {
    "match_all": {}
  },
  "sort": [
    { "field1": "asc" },
    { "field2": "desc" },
    { "field3": "desc" },
    { "field4": "asc" }
  ]
}'
```
```
[2024-07-30T15:44:31,838][INFO ][stdout                   ] [integTest-0] match_all []
[2024-07-30T15:44:31,838][INFO ][stdout                   ] [integTest-0] sort:
[2024-07-30T15:44:31,839][INFO ][stdout                   ] [integTest-0]   asc [field1]
[2024-07-30T15:44:31,839][INFO ][stdout                   ] [integTest-0]   asc [field4]
[2024-07-30T15:44:31,839][INFO ][stdout                   ] [integTest-0]   desc [field2]
[2024-07-30T15:44:31,839][INFO ][stdout                   ] [integTest-0]   desc [field3]
```

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
@@ -1,0 +1,322 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.plugin.insights.core.service.categorizer;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import org.opensearch.core.common.io.stream.NamedWriteable;
+import org.opensearch.index.query.AbstractGeometryQueryBuilder;
+import org.opensearch.index.query.CommonTermsQueryBuilder;
+import org.opensearch.index.query.ExistsQueryBuilder;
+import org.opensearch.index.query.FieldMaskingSpanQueryBuilder;
+import org.opensearch.index.query.FuzzyQueryBuilder;
+import org.opensearch.index.query.GeoBoundingBoxQueryBuilder;
+import org.opensearch.index.query.GeoDistanceQueryBuilder;
+import org.opensearch.index.query.GeoPolygonQueryBuilder;
+import org.opensearch.index.query.MatchBoolPrefixQueryBuilder;
+import org.opensearch.index.query.MatchPhrasePrefixQueryBuilder;
+import org.opensearch.index.query.MatchPhraseQueryBuilder;
+import org.opensearch.index.query.MatchQueryBuilder;
+import org.opensearch.index.query.MultiTermQueryBuilder;
+import org.opensearch.index.query.PrefixQueryBuilder;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.RangeQueryBuilder;
+import org.opensearch.index.query.RegexpQueryBuilder;
+import org.opensearch.index.query.SpanNearQueryBuilder;
+import org.opensearch.index.query.SpanTermQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
+import org.opensearch.index.query.TermsQueryBuilder;
+import org.opensearch.index.query.WildcardQueryBuilder;
+import org.opensearch.search.aggregations.AggregationBuilder;
+import org.opensearch.search.aggregations.AggregatorFactories;
+import org.opensearch.search.aggregations.PipelineAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.histogram.AutoDateHistogramAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.histogram.DateHistogramAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.histogram.HistogramAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.histogram.VariableWidthHistogramAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.missing.MissingAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.range.AbstractRangeBuilder;
+import org.opensearch.search.aggregations.bucket.range.GeoDistanceAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.range.IpRangeAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.sampler.DiversifiedAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.terms.RareTermsAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.terms.SignificantTermsAggregationBuilder;
+import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.AvgAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.CardinalityAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.ExtendedStatsAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.GeoCentroidAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.MaxAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.MinAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.StatsAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.SumAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.ValueCountAggregationBuilder;
+import org.opensearch.search.aggregations.support.ValuesSourceAggregationBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
+import org.opensearch.search.sort.SortBuilder;
+
+/**
+ * Class to generate query shape
+ */
+public class QueryShapeGenerator {
+    static final String EMPTY_STRING = "";
+    static final String ONE_SPACE_INDENT = " ";
+    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = FieldDataMapHelper.getQueryFieldDataMap();
+    static final Map<Class<?>, List<Function<Object, String>>> AGG_FIELD_DATA_MAP = FieldDataMapHelper.getAggFieldDataMap();
+    static final Map<Class<?>, List<Function<Object, String>>> SORT_FIELD_DATA_MAP = FieldDataMapHelper.getSortFieldDataMap();
+
+    /**
+     * Method to build search query shape given a source
+     * @param source search request source
+     * @param showFields whether to append field data
+     * @return Search query shape as String
+     */
+    public static String buildShape(SearchSourceBuilder source, Boolean showFields) {
+        StringBuilder shape = new StringBuilder();
+        shape.append(buildQueryShape(source.query(), showFields));
+        shape.append(buildAggregationShape(source.aggregations(), showFields));
+        shape.append(buildSortShape(source.sorts(), showFields));
+        return shape.toString();
+    }
+
+    /**
+     * Method to build query-section shape
+     * @param queryBuilder search request query builder
+     * @param showFields whether to append field data
+     * @return Query-section shape as String
+     */
+    static String buildQueryShape(QueryBuilder queryBuilder, Boolean showFields) {
+        if (queryBuilder == null) {
+            return EMPTY_STRING;
+        }
+        QueryShapeVisitor shapeVisitor = new QueryShapeVisitor();
+        queryBuilder.visit(shapeVisitor);
+        return shapeVisitor.prettyPrintTree(EMPTY_STRING, showFields);
+    }
+
+    /**
+     * Method to build aggregation shape
+     * @param aggregationsBuilder search request aggregation builder
+     * @param showFields whether to append field data
+     * @return Aggregation shape as String
+     */
+    static String buildAggregationShape(AggregatorFactories.Builder aggregationsBuilder, Boolean showFields) {
+        if (aggregationsBuilder == null) {
+            return EMPTY_STRING;
+        }
+        StringBuilder aggregationShape = recursiveAggregationShapeBuilder(
+            aggregationsBuilder.getAggregatorFactories(),
+            aggregationsBuilder.getPipelineAggregatorFactories(),
+            new StringBuilder(),
+            new StringBuilder(),
+            showFields
+        );
+        return aggregationShape.toString();
+    }
+
+    static StringBuilder recursiveAggregationShapeBuilder(
+        Collection<AggregationBuilder> aggregationBuilders,
+        Collection<PipelineAggregationBuilder> pipelineAggregations,
+        StringBuilder outputBuilder,
+        StringBuilder baseIndent,
+        Boolean showFields
+    ) {
+        //// Normal Aggregations ////
+        if (aggregationBuilders.isEmpty() == false) {
+            outputBuilder.append(baseIndent).append("aggregation:").append("\n");
+        }
+        List<String> aggShapeStrings = new ArrayList<>();
+        for (AggregationBuilder aggBuilder : aggregationBuilders) {
+            StringBuilder stringBuilder = new StringBuilder();
+            stringBuilder.append(baseIndent).append(ONE_SPACE_INDENT.repeat(2)).append(aggBuilder.getType());
+            if (showFields) {
+                stringBuilder.append(buildFieldDataString(AGG_FIELD_DATA_MAP.get(aggBuilder.getClass()), aggBuilder));
+            }
+            stringBuilder.append("\n");
+
+            if (aggBuilder.getSubAggregations().isEmpty() == false) {
+                // Recursive call on sub-aggregations
+                recursiveAggregationShapeBuilder(
+                    aggBuilder.getSubAggregations(),
+                    aggBuilder.getPipelineAggregations(),
+                    stringBuilder,
+                    baseIndent.append(ONE_SPACE_INDENT.repeat(4)),
+                    showFields
+                );
+                baseIndent.delete(0, 4);
+            }
+            aggShapeStrings.add(stringBuilder.toString());
+        }
+
+        // Sort alphanumerically and append aggregations list
+        Collections.sort(aggShapeStrings);
+        for (String shapeString : aggShapeStrings) {
+            outputBuilder.append(shapeString);
+        }
+
+        //// Pipeline Aggregation (cannot have sub-aggregations) ////
+        if (pipelineAggregations.isEmpty() == false) {
+            outputBuilder.append(baseIndent).append(ONE_SPACE_INDENT.repeat(2)).append("pipeline aggregation:").append("\n");
+
+            List<String> pipelineAggShapeStrings = new ArrayList<>();
+            for (PipelineAggregationBuilder pipelineAgg : pipelineAggregations) {
+                pipelineAggShapeStrings.add(
+                    new StringBuilder().append(baseIndent)
+                        .append(ONE_SPACE_INDENT.repeat(4))
+                        .append(pipelineAgg.getType())
+                        .append("\n")
+                        .toString()
+                );
+            }
+
+            // Sort alphanumerically and append pipeline aggregations list
+            Collections.sort(pipelineAggShapeStrings);
+            for (String shapeString : pipelineAggShapeStrings) {
+                outputBuilder.append(shapeString);
+            }
+        }
+        return outputBuilder;
+    }
+
+    /**
+     * Method to build sort shape
+     * @param sortBuilderList search request sort builders list
+     * @param showFields whether to append field data
+     * @return Sort shape as String
+     */
+    static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showFields) {
+        if (sortBuilderList == null || sortBuilderList.isEmpty()) {
+            return EMPTY_STRING;
+        }
+        StringBuilder sortShape = new StringBuilder();
+        sortShape.append("sort:\n");
+
+        List<String> shapeStrings = new ArrayList<>();
+        for (SortBuilder<?> sortBuilder : sortBuilderList) {
+            StringBuilder stringBuilder = new StringBuilder();
+            stringBuilder.append(ONE_SPACE_INDENT.repeat(2)).append(sortBuilder.order());
+            if (showFields) {
+                stringBuilder.append(buildFieldDataString(SORT_FIELD_DATA_MAP.get(sortBuilder.getClass()), sortBuilder));
+            }
+            shapeStrings.add(stringBuilder.toString());
+        }
+
+        Collections.sort(shapeStrings);
+        for (String line : shapeStrings) {
+            sortShape.append(line).append("\n");
+        }
+        return sortShape.toString();
+    }
+
+    /**
+     * Method to build field data
+     * @return String: comma separated list with leading space in square brackets
+     * Ex: " [my_field, width:5]"
+     */
+    static String buildFieldDataString(List<Function<Object, String>> methods, NamedWriteable builder) {
+        List<String> fieldDataList = new ArrayList<>();
+        if (methods != null) {
+            for (Function<Object, String> lambda : methods) {
+                fieldDataList.add(lambda.apply(builder));
+            }
+        }
+        return " [" + String.join(", ", fieldDataList) + "]";
+    }
+
+    /**
+     * Helper class to create static field data maps
+     */
+    private static class FieldDataMapHelper {
+
+        // Helper method to create map entries
+        private static <T> Map.Entry<Class<?>, List<Function<Object, String>>> createEntry(Class<T> clazz, Function<T, String> extractor) {
+            return Map.entry(clazz, List.of(obj -> extractor.apply(clazz.cast(obj))));
+        }
+
+        /**
+         * Returns a map where the keys are query builders, and the values are lists of
+         * functions that extract field values from instances of these classes.
+         *
+         * @return a map with class types as keys and lists of field extraction functions as values.
+         */
+        private static Map<Class<?>, List<Function<Object, String>>> getQueryFieldDataMap() {
+            return Map.ofEntries(
+                createEntry(AbstractGeometryQueryBuilder.class, AbstractGeometryQueryBuilder::fieldName),
+                createEntry(CommonTermsQueryBuilder.class, CommonTermsQueryBuilder::fieldName),
+                createEntry(ExistsQueryBuilder.class, ExistsQueryBuilder::fieldName),
+                createEntry(FieldMaskingSpanQueryBuilder.class, FieldMaskingSpanQueryBuilder::fieldName),
+                createEntry(FuzzyQueryBuilder.class, FuzzyQueryBuilder::fieldName),
+                createEntry(GeoBoundingBoxQueryBuilder.class, GeoBoundingBoxQueryBuilder::fieldName),
+                createEntry(GeoDistanceQueryBuilder.class, GeoDistanceQueryBuilder::fieldName),
+                createEntry(GeoPolygonQueryBuilder.class, GeoPolygonQueryBuilder::fieldName),
+                createEntry(MatchBoolPrefixQueryBuilder.class, MatchBoolPrefixQueryBuilder::fieldName),
+                createEntry(MatchQueryBuilder.class, MatchQueryBuilder::fieldName),
+                createEntry(MatchPhraseQueryBuilder.class, MatchPhraseQueryBuilder::fieldName),
+                createEntry(MatchPhrasePrefixQueryBuilder.class, MatchPhrasePrefixQueryBuilder::fieldName),
+                createEntry(MultiTermQueryBuilder.class, MultiTermQueryBuilder::fieldName),
+                createEntry(PrefixQueryBuilder.class, PrefixQueryBuilder::fieldName),
+                createEntry(RangeQueryBuilder.class, RangeQueryBuilder::fieldName),
+                createEntry(RegexpQueryBuilder.class, RegexpQueryBuilder::fieldName),
+                createEntry(SpanNearQueryBuilder.SpanGapQueryBuilder.class, SpanNearQueryBuilder.SpanGapQueryBuilder::fieldName),
+                createEntry(SpanTermQueryBuilder.class, SpanTermQueryBuilder::fieldName),
+                createEntry(TermQueryBuilder.class, TermQueryBuilder::fieldName),
+                createEntry(TermsQueryBuilder.class, TermsQueryBuilder::fieldName),
+                createEntry(WildcardQueryBuilder.class, WildcardQueryBuilder::fieldName)
+            );
+        }
+
+        /**
+         * Returns a map where the keys are aggregation builders, and the values are lists of
+         * functions that extract field values from instances of these classes.
+         *
+         * @return a map with class types as keys and lists of field extraction functions as values.
+         */
+        private static Map<Class<?>, List<Function<Object, String>>> getAggFieldDataMap() {
+            return Map.ofEntries(
+                createEntry(IpRangeAggregationBuilder.class, IpRangeAggregationBuilder::field),
+                createEntry(AutoDateHistogramAggregationBuilder.class, AutoDateHistogramAggregationBuilder::field),
+                createEntry(DateHistogramAggregationBuilder.class, DateHistogramAggregationBuilder::field),
+                createEntry(HistogramAggregationBuilder.class, HistogramAggregationBuilder::field),
+                createEntry(VariableWidthHistogramAggregationBuilder.class, VariableWidthHistogramAggregationBuilder::field),
+                createEntry(MissingAggregationBuilder.class, MissingAggregationBuilder::field),
+                createEntry(AbstractRangeBuilder.class, AbstractRangeBuilder::field),
+                createEntry(GeoDistanceAggregationBuilder.class, GeoDistanceAggregationBuilder::field),
+                createEntry(DiversifiedAggregationBuilder.class, DiversifiedAggregationBuilder::field),
+                createEntry(RareTermsAggregationBuilder.class, RareTermsAggregationBuilder::field),
+                createEntry(SignificantTermsAggregationBuilder.class, SignificantTermsAggregationBuilder::field),
+                createEntry(TermsAggregationBuilder.class, TermsAggregationBuilder::field),
+                createEntry(AvgAggregationBuilder.class, AvgAggregationBuilder::field),
+                createEntry(CardinalityAggregationBuilder.class, CardinalityAggregationBuilder::field),
+                createEntry(ExtendedStatsAggregationBuilder.class, ExtendedStatsAggregationBuilder::field),
+                createEntry(GeoCentroidAggregationBuilder.class, GeoCentroidAggregationBuilder::field),
+                createEntry(MaxAggregationBuilder.class, MaxAggregationBuilder::field),
+                createEntry(MinAggregationBuilder.class, MinAggregationBuilder::field),
+                createEntry(StatsAggregationBuilder.class, StatsAggregationBuilder::field),
+                createEntry(SumAggregationBuilder.class, SumAggregationBuilder::field),
+                createEntry(ValueCountAggregationBuilder.class, ValueCountAggregationBuilder::field),
+                createEntry(ValuesSourceAggregationBuilder.class, ValuesSourceAggregationBuilder::field)
+            );
+        }
+
+        /**
+         * Returns a map where the keys are sort builders, and the values are lists of
+         * functions that extract field values from instances of these classes.
+         *
+         * @return a map with class types as keys and lists of field extraction functions as values.
+         */
+        private static Map<Class<?>, List<Function<Object, String>>> getSortFieldDataMap() {
+            return Map.ofEntries(createEntry(FieldSortBuilder.class, FieldSortBuilder::getFieldName));
+        }
+    }
+}
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
@@ -29,7 +29,7 @@ import org.opensearch.telemetry.metrics.tags.Tags;
  */
 public final class SearchQueryCategorizer {
 
-    private static final Logger log = LogManager.getLogger(SearchQueryCategorizer.class);
+    private static final Logger logger = LogManager.getLogger(SearchQueryCategorizer.class);
 
     /**
      * Contains all the search query counters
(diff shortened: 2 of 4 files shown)


deshsidd:
Thanks! @dzane17 do you have some examples where we have the entire query shape including Query, Aggregations, Sort and Pipeline aggregations all in one?

Something like this:
```
bool
  filter:
    terms [field1:text]
  must:
    range [field2:date]
    term [field3:keyword]
  must_not:
    exists [field4:keyword]
  should:
    match [field5:text]
sort:
  asc [field6:number]
  desc [field2:date]
aggregation:
  terms [field3:keyword]
    aggregation:
      avg [field6:number]
      cardinality [field3:keyword]
      date_histogram [field2:date]
      max [field6:number]
      min [field6:number]
      percentile_ranks [field6:number]
      sum [field6:number]
pipeline_aggregation:
  avg_bucket
  max_bucket
```

Also, will the field related information for query operations will be as a followup?

deshsidd:
Was hoping we could have a `QueryShapeService`/`QueryShapeBuilder` or something similar that hides all the abstractions of processing the query shape. It only takes in an input search query source and logs the query shape. It can also include the string representation and the hash of a query shape. We can configure the builder to `include_field_details = True/False` and so on. 
Other services such as TopN and Categorization can now use this builder/service to get the query shape, log the query shape, get the unique hash for a query shape and so on. Let me know your thoughts @dzane17

ansjcy:
> Was hoping we could have a QueryShapeService/QueryShapeBuilder or something similar that hides all the abstractions of processing the query shape

+1 on this, we should create a lib to extract shape from a `source` obj so that we can use it in other places as well.

dzane17:
@deshsidd @ansjcy I have added a large search request example to the description.

QueryShapeService now contains logic to generate shape given a source. We can call it when and where ever needed. Currently all methods are static. I could not envision why we'd want to create instances of QueryShapeService unless we are calling `QueryShapeService.buildShape()` repeatedly for the same search. Let me know what you think. Also added unit tests for QueryShapeService.

@jainankitk Regarding query field data, there is not an easy way to extract field names from a `QueryBuilder` object. [This file QueryBuilders.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryBuilders.java) has a (complete?) list of classes which extend QueryBuilder. Some child classes have a `fieldName` field, some don't.

The options I came up with are 
1. Type checking - this is tedious, won't support new query types in the future, can break if existing query types change
2. Method overriding - update QueryBuilder parent class to return some blank value for fieldName() unless overridden in the child class
3. Ignore query field names completely

This also gets more complicated if we want to expand beyond field names... For example, we have `width` param which is unique to range type.
```
range [field2:value2, width:100]
```

deshsidd:
Thanks @dzane17.
For the large query shape why is `desc [sales_per_month>key]` in the shape?

Looks good overall, but need to figure out the best way to get the field details for the query type. Based on the above proposed approaches I like approach 2 `Method overriding - update QueryBuilder parent class to return some blank value for fieldName() unless overridden in the child class` @jainankitk let us know your thoughts here.

For properties such as width, we would have to do custom handling for the specific query type. Eg: if we encounter a range we can use the source to get the width.

jainankitk:
> @jainankitk Regarding query field data, there is not an easy way to extract field names from a QueryBuilder object. [This file QueryBuilders.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryBuilders.java) has a (complete?) list of classes which extend QueryBuilder. Some child classes have a fieldName field, some don't.

The options I came up with are

    Type checking - this is tedious, won't support new query types in the future, can break if existing query types change
    Method overriding - update QueryBuilder parent class to return some blank value for fieldName() unless overridden in the child class
    Ignore query field names completely

This also gets more complicated if we want to expand beyond field names... For example, we have width param which is unique to range type.

@dzane17 - Thank you for explaining the issue here, and some of the possible options. While option 2 looks decent, I am wondering about a fourth option to have a map containing QueryBuilder class type and list of things you want to collect for that specific class. For example:

<RangeQueryBuilder, List.of(getFieldName, getWidth)>
<TermBuilder, List.of(getFieldName)>

If a specific class is not present in the map, we will collect default data without any field names. Thoughts?

jainankitk:
> @jainankitk Regarding query field data, there is not an easy way to extract field names from a `QueryBuilder` object. [This file QueryBuilders.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryBuilders.java) has a (complete?) list of classes which extend QueryBuilder. Some child classes have a `fieldName` field, some don't.
> 
> The options I came up with are
> 
>     1. Type checking - this is tedious, won't support new query types in the future, can break if existing query types change
> 
>     2. Method overriding - update QueryBuilder parent class to return some blank value for fieldName() unless overridden in the child class
> 
>     3. Ignore query field names completely
> 
> 
> This also gets more complicated if we want to expand beyond field names... For example, we have `width` param which is unique to range type.
> 
> ```
> range [field2:value2, width:100]
> ```

@dzane17 - Thank you for explaining the issue here, and some of the possible options. While option 2 looks decent, I am wondering about a fourth option to have a map containing QueryBuilder class type and list of things you want to collect for that specific class. For example:

<RangeQueryBuilder, List.of(getFieldName, getWidth)>
<TermBuilder, List.of(getFieldName)>

If a specific class is not present in the map, we will collect default data without any field names. Thoughts?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeService.java
```diff
@@ -0,0 +1,142 @@
...
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
+import org.opensearch.search.sort.SortBuilder;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+public class QueryShapeService {
```
jainankitk, line 25: Do we really need to call this Service? Should be some sort of library without any lifecycle like `QueryShapeGenerator` or something?

dzane17:
@deshsidd `desc [sales_per_month>key]` is one of the sort criteria
```
"sort": [
    { "sales_per_month>key": { "order": "desc" } },
    { "total_sales": { "order": "desc" } },
    { "avg_order_quantity": { "order": "asc" } }
  ]
```

dzane17:
> > @jainankitk Regarding query field data, there is not an easy way to extract field names from a `QueryBuilder` object. [This file QueryBuilders.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/QueryBuilders.java) has a (complete?) list of classes which extend QueryBuilder. Some child classes have a `fieldName` field, some don't.
> > The options I came up with are
> > ```
> > 1. Type checking - this is tedious, won't support new query types in the future, can break if existing query types change
> > 
> > 2. Method overriding - update QueryBuilder parent class to return some blank value for fieldName() unless overridden in the child class
> > 
> > 3. Ignore query field names completely
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This also gets more complicated if we want to expand beyond field names... For example, we have `width` param which is unique to range type.
> > ```
> > range [field2:value2, width:100]
> > ```
> 
> @dzane17 - Thank you for explaining the issue here, and some of the possible options. While option 2 looks decent, I am wondering about a fourth option to have a map containing QueryBuilder class type and list of things you want to collect for that specific class. For example:
> 
> <RangeQueryBuilder, List.of(getFieldName, getWidth)> <TermBuilder, List.of(getFieldName)>
> 
> If a specific class is not present in the map, we will collect default data without any field names. Thoughts?

This 4th option will also work and solves additional field data beyond field names. The main disadvantage is that we need to maintain the Map.

If we intend to collect field data unique to each QueryBuilder, option 4 seems good. If we only need fieldNames, then option 2 seems more natural.

jainankitk:
> @jainankitk This 4th option will also work and solves additional field data beyond field names. The main disadvantage is that we need to maintain the Map.
> 
> If we intend to collect field data unique to each QueryBuilder, option 4 seems good. If we only need fieldNames, then option 2 seems more natural.

We can build the map gradually over time. We should write code in a manner that not having entry in the map still allows us to collect the basic information. As and when more stuff is added to the map, shape gets decorated with more stuff

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeService.java
dzane17: Changed to QueryShapeGenerator

dzane17:
@jainankitk @deshsidd I have added query field data using option 4 `Map<Class<?>, List<Function<Object, String>>>`. I have added all QueryBuilder classes which have a `fieldName()` method to the map. The List data structure is set up to accommodate additional field data in the future. Example in PR description is updated with new shape output.

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/QueryShapeGeneratorTests.java
```diff
@@ -0,0 +1,106 @@
...
+import org.opensearch.index.query.RegexpQueryBuilder;
+import org.opensearch.index.query.TermQueryBuilder;
+import org.opensearch.plugin.insights.core.service.categorizer.QueryShapeGenerator;
+import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
+import org.opensearch.search.aggregations.support.ValueType;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.SortOrder;
+import org.opensearch.test.OpenSearchTestCase;
+
+public final class QueryShapeGeneratorTests extends OpenSearchTestCase {
```
deshsidd, line 24: Should we add some tests for queries containing all 4 components in one query?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+public class QueryShapeGenerator {
+    static final String TWO_SPACE_INDENT = "  ";
+    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = Map.ofEntries(
+        Map.entry(AbstractGeometryQueryBuilder.class, List.of(obj -> ((AbstractGeometryQueryBuilder) obj).fieldName())),
```
deshsidd, line 50: nit: Can have a method to populate this to make the constructor look cleaner?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
```diff
@@ -84,11 +84,12 @@ public void categorize(SearchQueryRecord record) {
...
         Map<MetricType, Number> measurements = record.getMeasurements();
 
-        QueryBuilder topLevelQueryBuilder = source.query();
-        logQueryShape(topLevelQueryBuilder);
-        incrementQueryTypeCounters(topLevelQueryBuilder, measurements);
+        incrementQueryTypeCounters(source.query(), measurements);
         incrementQueryAggregationCounters(source.aggregations(), measurements);
         incrementQuerySortCounters(source.sorts(), measurements);
+
+        String searchShape = QueryShapeGenerator.buildShape(source, true);
```
deshsidd, line 91: Call build shape only when trace logging enabled

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/QueryShapeGeneratorTests.java
```diff
@@ -0,0 +1,106 @@
...
+        sourceBuilder.sort("album", SortOrder.ASC);
+
+        String shapeShowFieldsTrue = QueryShapeGenerator.buildShape(sourceBuilder, true);
+        String expectedShowFieldsTrue = "sort:\n" + "  asc [album]\n" + "  asc [price]\n" + "  desc [color]\n" + "  desc [vendor]\n";
+        assertEquals(expectedShowFieldsTrue, shapeShowFieldsTrue);
+
+        String shapeShowFieldsFalse = QueryShapeGenerator.buildShape(sourceBuilder, false);
+        String expectedShowFieldsFalse = "sort:\n" + "  asc\n" + "  asc\n" + "  desc\n" + "  desc\n";
+        assertEquals(expectedShowFieldsFalse, shapeShowFieldsFalse);
+    }
```
deshsidd, line 105: Can add tests for:
1. querybuilders not present in the field map or field missing
2. tests containing pipeline aggregations
3. large complex query shapes with nested indentation
4. making sure the normalization is working properly in terms of sorting of field_names and query operations, etc

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+        return shapeVisitor.prettyPrintTree("", showFields);
+    }
+
+    static String buildAggregationShape(AggregatorFactories.Builder aggregationsBuilder, Boolean showFields) {
+        if (aggregationsBuilder == null) {
+            return "";
+        }
+        StringBuilder aggregationShape = recursiveAggregationShapeBuilder(
+            aggregationsBuilder.getAggregatorFactories(),
+            aggregationsBuilder.getPipelineAggregatorFactories(),
```
deshsidd, line 105: Do pipeline agg need to be included in the recursive call? Can first take care of aggregations since they are recursive in nature and then handle the pipeline aggregations outside this recursion loop ?

deshsidd:
@jainankitk Let us know your thoughts

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+        //// Normal Aggregations ////
+        if (aggregationBuilders.isEmpty() == false) {
+            outputBuilder.append(baseIndent).append("aggregation:").append("\n");
+        }
+        List<String> aggShapeStrings = new ArrayList<>();
+        for (AggregationBuilder agg : aggregationBuilders) {
+            StringBuilder stringBuilder = new StringBuilder();
+            stringBuilder.append(baseIndent).append(TWO_SPACE_INDENT).append(agg.getType());
+            if (showFields) {
+                if (agg instanceof ValuesSourceAggregationBuilder) {
```
deshsidd, line 131: Don't need a map for agg similar to queries right?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+                    stringBuilder.append(" [").append(((ValuesSourceAggregationBuilder<?>) agg).field()).append("]");
+                } else {
+                    stringBuilder.append(" []");
+                }
+            }
+            stringBuilder.append("\n");
+
+            if (agg.getSubAggregations().isEmpty() == false) {
```
deshsidd, line 139: Do we need to do this in every recursive loop?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: My understanding is that pipeline aggregations can be a children of any normal aggregation. If a nested aggregation has a pipeline aggregation we want the shape to show that relationship instead of grouping all pipeline aggregations at the end.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Let me see if that can be done. That will make the code more extensible.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,195 @@
...
+import org.opensearch.search.aggregations.AggregatorFactories;
+import org.opensearch.search.aggregations.PipelineAggregationBuilder;
+import org.opensearch.search.aggregations.support.ValuesSourceAggregationBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
+import org.opensearch.search.sort.SortBuilder;
+
+public class QueryShapeGenerator {
+    static final String TWO_SPACE_INDENT = "  ";
+    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = Map.ofEntries(
```
jainankitk: Can we use `QueryBuilder` instead of `Object`?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,195 @@
...
+        Map.entry(AbstractGeometryQueryBuilder.class, List.of(obj -> ((AbstractGeometryQueryBuilder) obj).fieldName())),
+        Map.entry(CommonTermsQueryBuilder.class, List.of(obj -> ((CommonTermsQueryBuilder) obj).fieldName())),
+        Map.entry(org.opensearch.index.query.ExistsQueryBuilder.class, List.of(obj -> ((ExistsQueryBuilder) obj).fieldName())),
+        Map.entry(
+            org.opensearch.index.query.FieldMaskingSpanQueryBuilder.class,
+            List.of(obj -> ((FieldMaskingSpanQueryBuilder) obj).fieldName())
+        ),
```
jainankitk: nit: import these classes for readability

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Done

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
dzane17: Added check

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Good idea

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/QueryShapeGeneratorTests.java
dzane17: Added testComplexSearch()

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: Also, `Class<? extends QueryBuilder>`?

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/QueryShapeGeneratorTests.java
dzane17: 1. `BoolQueryBuilder` -> `bool []` is present. Plus added a few other aggs
2. Updated existing agg test
3. Added testComplexSearch()
4. This is tested

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,276 @@
...
+            }
+        }
+        return String.format(" [%s]", String.join(", ", fieldDataList));
+    }
+
+    /**
+     * Class to create field data map for query, agg, sort
+     * Map
+     */
+    public static class FieldDataMapHelper {
```
jainankitk: Why are we calling this FieldData Map? It can contain other information like query range, etc going forward

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,276 @@
...
+
+        // Helper method to create map entries
+        private static <T> Map.Entry<Class<?>, List<Function<Object, String>>> createEntry(Class<T> clazz, Function<T, String> extractor) {
+            return Map.entry(clazz, List.of(obj -> extractor.apply(clazz.cast(obj))));
+        }
+
+        // Method to return the QUERY_FIELD_DATA_MAP
+        public static Map<Class<?>, List<Function<Object, String>>> getQueryFieldDataMap() {
+            return Map.ofEntries(
+                createEntry(AbstractGeometryQueryBuilder.class, AbstractGeometryQueryBuilder::fieldName),
```
jainankitk, line 243: This should be stored as immutable map? And we return the reference everytime. This repetitive initialization can really slow down the search execution

jainankitk: Okay, I see it is invoked only once from outer class. Can we make these methods and the inner class private?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,276 @@
...
+    static final String TWO_SPACE_INDENT = "  ";
+    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = FieldDataMapHelper.getQueryFieldDataMap();
+    static final Map<Class<?>, List<Function<Object, String>>> AGG_FIELD_DATA_MAP = FieldDataMapHelper.getAggFieldDataMap();
+    static final Map<Class<?>, List<Function<Object, String>>> SORT_FIELD_DATA_MAP = FieldDataMapHelper.getSortFieldDataMap();
+
+    public static String buildShape(SearchSourceBuilder source, Boolean showFields) {
+        StringBuilder shape = new StringBuilder();
+        shape.append(buildQueryShape(source.query(), showFields));
+        shape.append(buildAggregationShape(source.aggregations(), showFields));
+        shape.append(buildSortShape(source.sorts(), showFields));
```
jainankitk, line 87: are we planning to add the logic for anonymizing fieldnames? to field1, field2, field3 and so on?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: I cannot directly make this change. Query, agg, sort all use the same helper method so we'd need 3 version of `createEntry()` for QueryBuilder, AggBuilder, and SortBuilder. I think this can work, but will cost some extra lines.

```
        private static <T> Map.Entry<Class<?>, List<Function<Object, String>>> createEntry(Class<T> clazz, Function<T, String> extractor) {
            return Map.entry(clazz, List.of(obj -> extractor.apply(clazz.cast(obj))));
        }
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Done, my IDE changed it automatically somehow.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Yes, maps are only created once. Changed `FieldDataMapHelper` and its methods to private

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,318 @@
...
+            }
+            stringBuilder.append("\n");
+
+            if (aggBuilder.getSubAggregations().isEmpty() == false) {
+                // Recursive call on sub-aggregations
+                recursiveAggregationShapeBuilder(
+                    aggBuilder.getSubAggregations(),
+                    aggBuilder.getPipelineAggregations(),
+                    stringBuilder,
+                    indentCount + 2,
```
jainankitk, line 156: Is the indent 4 or two spaces? Also, having repeating one space indent might be more efficient, as the repeat method can use below code path:

```
if (len == 1) {
        final byte[] single = new byte[count];
        Arrays.fill(single, value[0]);
        return new String(single, coder);
}
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,318 @@
...
+                    showFields
+                );
+            }
+            aggShapeStrings.add(stringBuilder.toString());
+        }
+
+        // Sort alphanumerically and append aggregations list
```
jainankitk, line 163: Just to clarify, this is sorting between 2 aggregations at the same level, right?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,318 @@
...
+            for (PipelineAggregationBuilder pipelineAgg : pipelineAggregations) {
+                pipelineAggShapeStrings.add(baseIndent + TWO_SPACE_INDENT.repeat(2) + pipelineAgg.getType() + "\n");
+            }
+
+            // Sort alphanumerically and append pipeline aggregations list
+            Collections.sort(pipelineAggShapeStrings);
+            for (String shapeString : pipelineAggShapeStrings) {
+                outputBuilder.append(shapeString);
+            }
+        }
```
jainankitk, line 183: This logic looks repetitive. can we not invoke the recursive function twice, once for default and once for pipeline aggregations? Even if pipelines don't have subaggregation, the recursive invocation will always be skipped!

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Yes, sorting is only needed for children with same parent.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: I was not planning to do so in this PR. @deshsidd do we have anonymization in the roadmap?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: We are using field data as a general term which contains field name, type, range, etc... I can change it to something like attributes or properties but can't think of a great name atm.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Indent is 2 spaces. It is incremented by 2 here because every level has a header "aggregation:" followed by its details "  terms [field]" so the next level is indented by 2. 

I changed it to 1 space indent although I could not find any specific info on performance impact of repeat() for string length 1 vs 2...

dzane17:
@jainankitk @deshsidd Currently we have no field data for pipeline aggregations. Should we use `bucketsPaths`?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Yes there is some repeated structure but using the same recursive function will not work. Headers and field data are handled differently so we'd need to check for AggregationBuilder or PipelineAggregationBuilder, cast the object, then evaluate accordingly. Given the amount if divergence it is simpler to keep in a single function. 

I also tried overloading but that causes "both methods have same erasure" error since both signatures use `Collection<X>`.
```
recursiveAggregationShapeBuilder(
        Collection<AggregationBuilder> aggregationBuilders,
        StringBuilder outputBuilder,
        int indentCount,
        Boolean showFields
    )
```

For either option there is no computational saving and will only increase the number of recursive branches.

jainankitk: Can we atleast have 2 separate methods for default and pipeline aggregations? Since they don't have anything to do with each other

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: This blog might be good read: https://medium.com/@ggajos/java-11-how-string-repeat-was-implemented-and-why-d93796b7abba

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,318 @@
...
+            }
+            stringBuilder.append("\n");
+
+            if (aggBuilder.getSubAggregations().isEmpty() == false) {
+                // Recursive call on sub-aggregations
+                recursiveAggregationShapeBuilder(
+                    aggBuilder.getSubAggregations(),
+                    aggBuilder.getPipelineAggregations(),
+                    stringBuilder,
+                    indentCount + 2,
```
jainankitk, line 156: This should be 4 with `ONE_SPACE_INDENT`

jainankitk: Also, the indent might be passed as StringBuilder or Buffer by adding indent characters on top, removing them at the bottom

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -0,0 +1,318 @@
...
+            outputBuilder.append(shapeString);
+        }
+
+        //// Pipeline Aggregation (cannot have sub-aggregations) ////
+        if (pipelineAggregations.isEmpty() == false) {
+            outputBuilder.append(baseIndent).append(ONE_SPACE_INDENT).append("pipeline aggregation:").append("\n");
+
+            List<String> pipelineAggShapeStrings = new ArrayList<>();
+            for (PipelineAggregationBuilder pipelineAgg : pipelineAggregations) {
+                pipelineAggShapeStrings.add(baseIndent + ONE_SPACE_INDENT.repeat(2) + pipelineAgg.getType() + "\n");
```
jainankitk, line 175: Use `StringBuilder` instead of `String::append` wherever possible.

jainankitk (approved):
Initial changes LGTM! Looking forward to the anonymization and using the hash of query shape, and exposing that through topN to begin with. cc: @ansjcy @deshsidd

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Changed to StringBuilder and fixed indent size

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Good point

deshsidd:
> @jainankitk @deshsidd Currently we have no field data for pipeline aggregations. Should we use `bucketsPaths`?

Not sure if this will be helpful. Can leave it without any additional data for now?

deshsidd:
> Initial changes LGTM! Looking forward to the anonymization and using the hash of query shape, and exposing that through topN to begin with. cc: @ansjcy @deshsidd

Yup, next would be to expose a function using hash of query shape and using it in the grouping top n by similarity feature.

deshsidd:
Thanks @dzane17 this is looking good overall! Lets iterate and keep improving this.

deshsidd (approved):

PR #66, opened 2024-08-02
Query grouping framework for Top N queries and group by query similarity 

deshsidd:
### Description
For Top N queries by latency, we can encounter scenarios where some (or most) of the Top N queries contain duplicate queries. Say the same dashboard query is triggered continuously and happens to be the most expensive query in terms of latency - in this scenario all the Top N queries by latency will likely be spammed by the same query. To overcome such scenarios and to get a more detailed view of the Top N query patterns we have implemented Grouping Top N queries by similarity. As a followup we can also use this framework to implement grouping top N queries by frequency, user_id, etc.

Major changes:
1. Query Grouping Service that groups queries based on a group_id and uses a Min and Max priority queue approach as discussed in the RFC
2. Created the Measurement class as an abstraction for number that is used to store the measurement for the specific MetricType. Measurement can support DimensionType (Average, Sum) for the specific measurement. For grouping by similarity we use the average latency, average cpu and average memory to maintain the ordering.
3. We have a GroupingType enum that describes how we group the Top N queries (similarity, user_id)
4. The Grouping setting applies to ALL metric types and we cannot set this only for a subset of MetricType as discussed in the RFC.
5. Each TopQueriesService has its instance of QueryGroupingService. We have one TopQueriesService for each metrictype.
6. In QueryInsightsService we add ALL the records to the queryRecordsQueue for the TopQueriesService to consume if search.query.metric feature is enabled or is grouping enabled. Note that we skip the optimization in this case.

```
public boolean addRecord(final SearchQueryRecord record) {
        boolean shouldAdd = isSearchQueryMetricsFeatureEnabled() || isGroupingEnabled();
        if (!shouldAdd) {
            for (Map.Entry<MetricType, TopQueriesService> entry : topQueriesServices.entrySet()) {
                if (!enableCollect.get(entry.getKey())) {
                    continue;
                }
                List<SearchQueryRecord> currentSnapshot = entry.getValue().getTopQueriesCurrentSnapshot();
                // skip add to top N queries store if the incoming record is smaller than the Nth record
                if (currentSnapshot.size() < entry.getValue().getTopNSize()
                    || SearchQueryRecord.compare(record, currentSnapshot.get(0), entry.getKey()) > 0) {
                    shouldAdd = true;
                    break;
                }
            }
        }
        if (shouldAdd) {
            return queryRecordsQueue.offer(record);
        }
        return false;
    }
```
7. Added exhaustive unit tests for QueryGroupingService.

### Issues Resolved
addresses [#13357](https://github.com/opensearch-project/OpenSearch/issues/13357)

### Configure Grouping
```
deshsid@c889f3bdacfb query-insights-unzip % curl -X PUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent": {
    "search.insights.top_queries.group_by": "similarity"
  }
}
'
{"acknowledged":true,"persistent":{"search":{"insights":{"top_queries":{"group_by":"similarity"}}}},"transient":{}}%
```
### Configure Grouping Error Response
```
curl -X PUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent": {
    "search.insights.top_queries.group_by": "similarit"
  }
}'

{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"illegal value can't update [search.insights.top_queries.group_by] from [similarity] to [similarit]"}],"type":"illegal_argument_exception","reason":"illegal value can't update [search.insights.top_queries.group_by] from [similarity] to [similarit]","caused_by":{"type":"illegal_argument_exception","reason":"Invalid grouping type [similarit], type should be one of [SIMILARITY, USER_ID, NONE]"}},"status":400}%
```

### Get Top N Queries by latency with grouping enabled, group_by SIMILARITY

```
curl -XGET "http://localhost:9200/_insights/top_queries"
{
  "top_queries": [
    {
      "timestamp": 1722630496342,
      "query_hashcode": 29791,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 135,
          "parentTaskId": 134,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 625000,
            "memory_in_bytes": 41512
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 134,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 84000,
            "memory_in_bytes": 3264
          }
        }
      ],
      "source": {},
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 774,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 774,
          "count": 1,
          "dimensionType": "AVERAGE"
        }
      }
    },
    {
      "timestamp": 1722630528201,
      "query_hashcode": 709023605,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 163,
          "parentTaskId": 162,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 9412000,
            "memory_in_bytes": 618968
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 162,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 158000,
            "memory_in_bytes": 3720
          }
        }
      ],
      "source": {
        "sort": [
          {
            "age": {
              "order": "asc"
            }
          }
        ]
      },
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 10,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 11,
          "count": 1,
          "dimensionType": "AVERAGE"
        }
      }
    },
    {
      "timestamp": 1722630499772,
      "query_hashcode": -1204891025,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 137,
          "parentTaskId": 136,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 7603000,
            "memory_in_bytes": 477600
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 136,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 127000,
            "memory_in_bytes": 3232
          }
        }
      ],
      "source": {
        "query": {
          "match": {
            "occupation": {
              "query": "Software Engineer",
              "operator": "OR",
              "prefix_length": 0,
              "max_expansions": 50,
              "fuzzy_transpositions": true,
              "lenient": false,
              "zero_terms_query": "NONE",
              "auto_generate_synonyms_phrase_query": true,
              "boost": 1.0
            }
          }
        }
      },
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 8,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 9,
          "count": 1,
          "dimensionType": "AVERAGE"
        }
      }
    }
  ]
}

```

### Get Top N queries with group_by NONE
```
{
  "top_queries": [
    {
      "timestamp": 1722632764895,
      "query_hashcode": -1204891025,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 953,
          "parentTaskId": 952,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 1640000,
            "memory_in_bytes": 120760
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 952,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 150000,
            "memory_in_bytes": 3232
          }
        }
      ],
      "source": {
        "query": {
          "match": {
            "occupation": {
              "query": "Software Engineer",
              "operator": "OR",
              "prefix_length": 0,
              "max_expansions": 50,
              "fuzzy_transpositions": true,
              "lenient": false,
              "zero_terms_query": "NONE",
              "auto_generate_synonyms_phrase_query": true,
              "boost": 1.0
            }
          }
        }
      },
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 2,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 3,
          "count": 1,
          "dimensionType": "NONE"
        }
      }
    },
    {
      "timestamp": 1722632770456,
      "query_hashcode": 605146258,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 959,
          "parentTaskId": 958,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 852000,
            "memory_in_bytes": 49328
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 958,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 120000,
            "memory_in_bytes": 3240
          }
        }
      ],
      "source": {
        "query": {
          "range": {
            "age": {
              "from": 30,
              "to": null,
              "include_lower": false,
              "include_upper": true,
              "boost": 1.0
            }
          }
        }
      },
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 1,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 2,
          "count": 1,
          "dimensionType": "NONE"
        }
      }
    },
    {
      "timestamp": 1722632769697,
      "query_hashcode": 605146258,
      "search_type": "query_then_fetch",
      "task_resource_usages": [
        {
          "action": "indices:data/read/search[phase/query]",
          "taskId": 957,
          "parentTaskId": 956,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 980000,
            "memory_in_bytes": 49328
          }
        },
        {
          "action": "indices:data/read/search",
          "taskId": 956,
          "parentTaskId": -1,
          "nodeId": "zp2vxuVsRwawzBK2u7f7FA",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 130000,
            "memory_in_bytes": 3240
          }
        }
      ],
      "source": {
        "query": {
          "range": {
            "age": {
              "from": 30,
              "to": null,
              "include_lower": false,
              "include_upper": true,
              "boost": 1.0
            }
          }
        }
      },
      "indices": ["my_test_index"],
      "total_shards": 1,
      "labels": {},
      "phase_latency_map": {
        "expand": 0,
        "query": 1,
        "fetch": 0
      },
      "node_id": "zp2vxuVsRwawzBK2u7f7FA",
      "measurements": {
        "latency": {
          "metricType": "latency",
          "number": 2,
          "count": 1,
          "dimensionType": "NONE"
        }
      }
    }
  ]
}
```

Uploading Screen Recording 2024-08-28 at 4.35.10 PM.mov…


src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -39,10 +40,11 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
      */
     public SearchQueryRecord(final StreamInput in) throws IOException, ClassCastException {
         this.timestamp = in.readLong();
-        measurements = new HashMap<>();
-        in.readMap(MetricType::readFromStream, StreamInput::readGenericValue)
-            .forEach(((metricType, o) -> measurements.put(metricType, metricType.parseValue(o))));
+        measurements = new LinkedHashMap<>();
```
ansjcy, line 43: Why it is needed to use `LinkedHashMap` ?

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -115,9 +136,12 @@ public XContentBuilder toXContent(final XContentBuilder builder, final ToXConten
         for (Map.Entry<Attribute, Object> entry : attributes.entrySet()) {
             builder.field(entry.getKey().toString(), entry.getValue());
         }
-        for (Map.Entry<MetricType, Number> entry : measurements.entrySet()) {
-            builder.field(entry.getKey().toString(), entry.getValue());
+        builder.startObject("measurements");
```
ansjcy, line 139: Please use constant for Object key

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -33,7 +34,7 @@ public class QueryInsightsSettings {
     /**
      * Max number of requests for the consumer to collect at one time
      */
-    public static final int QUERY_RECORD_QUEUE_CAPACITY = 1000;
+    public static final int QUERY_RECORD_QUEUE_CAPACITY = 100000;
```
ansjcy, line 37: How is this decided? Please do benchmark tests with some real-world workload to justify the capacity limit.

src/test/java/org/opensearch/plugin/insights/core/service/QueryGroupingServiceTests.java
```diff
@@ -0,0 +1,1041 @@
...
+        }
+
+        assertEquals(2, queryGroupingService.numberOfTopGroups());
+        assertEquals(1000L, topQueriesStore.poll().getMeasurement(MetricType.LATENCY));
+        assertEquals(1100L, topQueriesStore.poll().getMeasurement(MetricType.LATENCY));
+    }
+
+    // 7. Existing query group update to MAX increases average - stay in MAX
+    public void testExistingGroupUpdateToMaxIncreaseAverageStayInMax() {
+        queryGroupingService = new QueryGroupingService(
```
ansjcy, line 681: A lot of the code in those cases are the same set up. Please simplify with a shared function.

ansjcy (changes requested):
The overall class designs looks good to me.

But I have concerns on the correctness of the logic in QueryGroupingService to implement algorithms proposed in https://github.com/opensearch-project/OpenSearch/issues/13357#issuecomment-2157563513. Please see the individual comments for details.

Also, a lot of the heap operations are `O(n)` and `O(total possible number of groups)`, which is not acceptable. Please refer to the comment: https://github.com/opensearch-project/OpenSearch/issues/13357#issuecomment-2159162672 to resolve it.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+        return aggregateSearchQueryRecord;
+    }
+
+    private void addToMinPQ(SearchQueryRecord searchQueryRecord, String groupId) {
+        minHeapTopQueriesStore.add(searchQueryRecord);
+        groupIdToAggSearchQueryRecord.put(groupId, new Tuple<>(searchQueryRecord, true));
+
+        while (minHeapTopQueriesStore.size() > topNSize) {
```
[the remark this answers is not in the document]
deshsidd: This is to handle the case where the Top N size could have changed in the interim.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
```diff
@@ -0,0 +1,196 @@
...
+        minHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
+        Number measurementToAdd = searchQueryRecord.getMeasurement(metricType);
+        aggregateSearchQueryRecord.addMeasurement(metricType, measurementToAdd);
+
+        if (maxHeapQueryStore.size() > 0) {
+            addToMaxPQ(aggregateSearchQueryRecord, groupId);
+        } else {
+            addToMinPQ(aggregateSearchQueryRecord, groupId);
+        }
+    }
```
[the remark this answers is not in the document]
deshsidd: Please check the following [algorithm](https://github.com/opensearch-project/OpenSearch/issues/13357#issuecomment-2269706425) and let me know your thoughts.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
```diff
@@ -0,0 +1,196 @@
...
+     * @param searchQueryRecord record
+     * @return Grouping Id
+     */
+    private String getGroupingId(SearchQueryRecord searchQueryRecord) {
+        switch (groupingType) {
+            case SIMILARITY:
+                return searchQueryRecord.getAttributes().get(Attribute.QUERY_HASHCODE).toString();
+            case NONE:
+                throw new IllegalArgumentException("Should not try to group queries if grouping type is NONE");
+            default:
```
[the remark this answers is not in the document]
deshsidd: Will need to add support for USER_ID, let me remove it from the enum for now

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -114,15 +121,17 @@ public QueryInsightsService(
...
 
     /**
      * Ingest the query data into in-memory stores
      *
      * @param record the record to ingest
+     * @return true/false
      */
     public boolean addRecord(final SearchQueryRecord record) {
-        boolean shouldAdd = searchQueryMetricsEnabled;
+        boolean shouldAdd = isSearchQueryMetricsFeatureEnabled() || isGroupingEnabled();
```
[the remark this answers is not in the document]
deshsidd: This is because if grouping is enabled, we need to add all records to the queryRecordsQueue and avoid the following optimization in QueryInsightsService:
```
// skip add to top N queries store if the incoming record is smaller than the Nth record
                if (currentSnapshot.size() < entry.getValue().getTopNSize()
                    || SearchQueryRecord.compare(record, currentSnapshot.get(0), entry.getKey()) > 0) {
                    shouldAdd = true;
                    break;
                }
```

Let me know if you think otherwise

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -186,6 +195,40 @@ public void enableCollection(final MetricType metricType, final boolean enable)
         this.topQueriesServices.get(metricType).setEnabled(enable);
     }
 
+    /**
+     * Validate grouping
```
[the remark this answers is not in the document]
deshsidd: Need it to pass the code health checks?

src/main/java/org/opensearch/plugin/insights/rules/model/GroupingType.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Grouping type to group the Top N queries
+ */
+public enum GroupingType {
+    NONE("none"),
+    SIMILARITY("similarity"),
+    USER_ID("userid");
```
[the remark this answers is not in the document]
deshsidd: removing for now, will have a separate PR to onboard grouping by USER_ID

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Add measurement number to the current number based on the dimension type
+     * @param toAdd number to add
+     */
+    public void addMeasurement(Number toAdd) {
+        switch (dimensionType) {
+            case NONE:
+                dimensionType = DimensionType.SUM;
```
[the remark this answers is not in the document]
deshsidd: Makes sense, trying to think of any scenario where we might use the AggregationType = None.
If we cannot think of any will remove NONE and default to SUM.

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Get measurement number based on the dimension type
+     * @return measurement number
+     */
+    public Number getMeasurement() {
+        switch (dimensionType) {
+            case NONE:
+            case SUM:
```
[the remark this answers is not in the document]
deshsidd: Will remove NONE if we cannot think of any scenario where we might need it

src/main/java/org/opensearch/plugin/insights/rules/model/MetricType.java
```diff
@@ -97,13 +97,35 @@ public int compare(final Number a, final Number b) {
         return -1;
     }
 
+    public static Number addMeasurements(Number a, Number b, MetricType metricType) {
+        switch (metricType) {
+            case LATENCY:
+            case MEMORY:
+            case CPU:
+                return a.longValue() + b.longValue();
```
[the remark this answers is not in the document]
deshsidd: Yup good point, overflow can occur here. We can either consider using BigInteger or check for overflow:
```
public static Number addMeasurements(Number a, Number b, MetricType metricType) {
    switch (metricType) {
        case LATENCY:
        case MEMORY:
        case CPU:
            try {
                return Math.addExact(a.longValue(), b.longValue());
            } catch (ArithmeticException e) {
                throw new ArithmeticException("Overflow occurred while adding measurements: " + e.getMessage());
            }
        default:
            throw new IllegalArgumentException("Unsupported metric type: " + metricType);
    }
}
```
But not sure how to proceed on exception. Will have to think about this one. How likely is it that the CPU value will exceed the `long` limits?

The range of Long is [-9,223,372,036,854,775,808, +9,223,372,036,854,775,807].

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
deshsidd: Needed to make sure the writing and reading the object to IO works as expected. `testSerialize()` and some others were failing before this.

src/test/java/org/opensearch/plugin/insights/QueryInsightsTestUtils.java
```diff
@@ -57,41 +61,69 @@ public QueryInsightsTestUtils() {}
...
+            long latencyValue = randomLongBetween(1000, 10000); // Replace with actual method to generate a random long
+            long cpuValue = randomLongBetween(1000, 10000);
+            long memoryValue = randomLongBetween(1000, 10000);
+            Map<MetricType, Measurement> measurements = new LinkedHashMap<>();
+            measurements.put(MetricType.LATENCY, new Measurement(MetricType.LATENCY, latencyValue, dimensionType));
+            measurements.put(MetricType.CPU, new Measurement(MetricType.CPU, cpuValue, dimensionType));
+            measurements.put(MetricType.MEMORY, new Measurement(MetricType.MEMORY, memoryValue, dimensionType));
 
-            Map<String, Long> phaseLatencyMap = new HashMap<>();
+            Map<String, Long> phaseLatencyMap = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
deshsidd: testSerialize() failing due to ordering issues

deshsidd:
We can run some benchmarks to view the performance here and keep this feature as experimental/beta and also limit the number of groups. If needed we can use an [indexed priority queue](https://algs4.cs.princeton.edu/24pq/IndexMinPQ.java.html) from [here](https://mvnrepository.com/artifact/edu.princeton.cs/algs4/1.0.4) as followup changes. Let me know your thoughts.

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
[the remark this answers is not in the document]
deshsidd: Displaying dimension type = SUM on the API response might be misleading if group_by = NONE since its a singular value and not a sum.
"dimensionType": "NONE" might be more suitable?

ansjcy:
> If needed we can use an indexed priority queue as followup changes.

I don't think this is a good idea, the whole algorithm mentioned in https://github.com/opensearch-project/OpenSearch/issues/13357#issuecomment-2157563513 is based on using indexed pq to store the groups. Otherwise we are storing all the queries groups, updating / deletion can take O(total possible number of groups) in a worst case scenario, which is not acceptable.

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
```diff
@@ -0,0 +1,160 @@
...
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+
+/**
+ * Measurement that is stored in the SearchQueryRecord. Measurement can be of a specific DimensionType and MetricType
+ */
+public class Measurement implements ToXContentObject, Writeable {
+    private static int DEFAULT_COUNT = 1;
+    private DimensionType dimensionType;
+    private MetricType metricType;
```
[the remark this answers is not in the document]
deshsidd: Decoupled measurement and metrictype in the latest

deshsidd:
Made all the required refactoring based on the comments. Highlights include:
1. Decouple Measurement and MetricType
2. Ensure NONE aggregation type performs no aggregations.
3. Refactor unit tests to re-use code whenever applicable
4. Added one missing edge case in QueryGroupingService algorithm


Only major open question is regarding the java priority queue verses indexed priority queue.
- Indexed priority queue has O(logn) updates while with the java PQ it takes O(n) since we have to remove (O(n)) and then  re-add (O(logn)). Note that n is the number of groups in a Top N window.
- Only viable [indexed priority queue](https://algs4.cs.princeton.edu/24pq/IndexMinPQ.java.html) I found was from [here](https://mvnrepository.com/artifact/edu.princeton.cs/algs4/1.0.4). Tried including this in gradle and got errors due to the following [issue](https://github.com/kevin-wayne/algs4/issues/121). There seems to be [other ways](https://github.com/kevin-wayne/algs4/issues/102) to add this library but not sure we want to pursue these unconventional routes. Furthermore, not sure about the stability and community support for this library.
- We can also consider trying to limit the number of groups per window or implementing our own version of indexed PQ to improve the performance.

Lets discuss more and figure out a path forward!

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
```diff
@@ -0,0 +1,243 @@
...
+        // - If max PQ is empty, add to min PQ and overflow any records to max PQ
+        // 2. If present in max PQ
+        // - remove the record from the max PQ
+        // - update the aggregate record (aggregate measurement could increase or decrease)
+        // - If min PQ is full, add to min PQ and overflow any records to max PQ
+        // - else, add to max PQ and promote any records to min PQ
+        else {
+            aggregateSearchQueryRecord = groupIdToAggSearchQueryRecord.get(groupId).v1();
+            boolean isPresentInMinPQ = groupIdToAggSearchQueryRecord.get(groupId).v2();
+            if (isPresentInMinPQ) {
```
jainankitk, line 124: How are we doing the aggregation on measurement? Is it just the sum of all measurements or something else? Also, how are we adhering to the running window constraint if we aggregate the measurements together?

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -109,6 +113,13 @@ public class TopQueriesService {
         topQueriesStore = new PriorityQueue<>(topNSize, (a, b) -> SearchQueryRecord.compare(a, b, metricType));
         topQueriesCurrentSnapshot = new AtomicReference<>(new ArrayList<>());
         topQueriesHistorySnapshot = new AtomicReference<>(new ArrayList<>());
+        queryGroupingService = new QueryGroupingService(
+            metricType,
+            QueryInsightsSettings.DEFAULT_GROUPING_TYPE,
+            AggregationType.AVERAGE,
+            topQueriesStore,
+            topNSize
+        );
```
jainankitk, line 122: I don't think we should be calling this `QueryGroupingService`. Service generally acts independently through well defined interaction points. Maybe `QueryGrouper` is more appropriate term.

src/main/java/org/opensearch/plugin/insights/rules/model/AggregationType.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Aggregation type for a measurement. Default is NONE and average is used for grouping Top N queries by similarity.
+ */
+public enum AggregationType {
+    NONE,
+    AVERAGE,
+    SUM;
+
+    public static AggregationType DEFUALT_AGGREGATION_TYPE = NONE;
```
jainankitk, line 19: nit: typo

jainankitk (commented):
Overall the logic for query grouping seems complex, and I am wondering if there is a way to simplify some of it by making some reasonable assumptions

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
deshsidd: Its based on the AggregationType (NONE, AVG, SUM)

src/main/java/org/opensearch/plugin/insights/core/service/QueryGroupingService.java
```diff
@@ -0,0 +1,243 @@
...
+     * Min heap to keep track of the Top N query groups and is passed from TopQueriesService as the topQueriesStore
+     */
+    private PriorityQueue<SearchQueryRecord> minHeapTopQueriesStore;
+    /**
+     * The Max heap is an overflow data structure used to manage records that exceed the capacity of the Min heap.
+     * It stores all records not included in the Top N query results. When the aggregate measurement for one of these
+     * records is updated and it now qualifies as part of the Top N, the record is moved from the Max heap to the Min heap,
+     * and the records are rearranged accordingly.
+     */
+    private PriorityQueue<SearchQueryRecord> maxHeapQueryStore;
```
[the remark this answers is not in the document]
deshsidd: Yup. Running some benchmarks to see what is a reasonable assumption for max number of groups. Thinking of having a setting to set the max_number_of_groups and limit this PQ to that number. If we exceed this number we can drop and add debug logs

deshsidd:
Ran some benchmarks to figure out a reasonable number for the cardinality of the groups and here are the results:

1. logging heavy: `http_logs`
Here are the number of groups logged at the end of a window cycle over the course of approx 3 days:
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 8, 5, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 6, 7, 3, 7, 6, 3, 0, 0, 0, 0, 0, 0, 0, 0, 7, 6, 3, 0, 0, 0, 0, 0, 0, 0, 0, 7, 6, 3, 0]
Maximum: 8

2. search heavy: `nyc_taxis`
Here are the number of groups logged at the end of a window cycle over the course of approx 3 days:
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3, 6, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 8, 0, 0, 0, 8, 0, 0, 0, 0, 0, 0, 0, 0, 2, 7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 8, 0, 0]
Maximum: 8

3. custom workload simulating real world traffic:
[0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 2, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 7, 0, 0, 0, 0, 0, 0, 2, 0, 0, 1, 4, 0, 0, 0, 0, 0, 0]
Maximum: 7
 
Note that window size set to : `1 hour`

IMHO it might be reasonable to having a setting `max_groups` to set the maximum number of groups and limit the PQ to that number. If we exceed this number we can drop and add debug logs. The `max_groups` value should have a validation such that it cannot be set beyond 10,000.

dzane17:
Is it possible we introduce other "group by" attributes in the future? I am wondering if we can simplify the cluster setting from ["none", "similarity"] to a boolean.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -185,6 +194,62 @@ public void enableCollection(final MetricType metricType, final boolean enable)
...
+            this.topQueriesServices.get(metricType).setMaxGroups(maxGroups);
+        }
+    }
+
+    /**
+     * Validate max number of groups. Should be between 1 and MAX_GROUPS_LIMIT
+     * @param maxGroups maximum number of groups that should be tracked when calculating Top N groups
+     */
+    public void validateMaximumGroups(final int maxGroups) {
+        if (maxGroups < 1 || maxGroups > QueryInsightsSettings.MAX_GROUPS_LIMIT) {
```
dzane17, line 237: If `maxGroups` < N it can lead to unexpected results. Should we validate that here?

deshsidd: maxGroups only kicks in after the minimum queue is filled. So first we fill in the min queue for Top N. We then start overflowing groups to the max queue. The maxGroups will limit the size of the max queue.

dzane17 (approved):

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+public class QueryGrouper {
+
+    /**
+     * Logger
+     */
+    private static final Logger log = LogManager.getLogger(QueryGrouper.class);
+    /**
+     * Grouping type for the current grouping service
+     */
+    private GroupingType groupingType;
```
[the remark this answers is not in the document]
deshsidd: Why does this need to be volatile? `queryGrouper` is initialized in `TopQueriesService` and called only from that service. We have a different queryGrouper for each TopQueryService (one for each metricType).

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+import org.opensearch.plugin.insights.rules.model.AggregationType;
+import org.opensearch.plugin.insights.rules.model.Attribute;
+import org.opensearch.plugin.insights.rules.model.GroupingType;
+import org.opensearch.plugin.insights.rules.model.MetricType;
+import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
+import org.opensearch.plugin.insights.settings.QueryInsightsSettings;
+
+/**
+ * Handles grouping of search queries based on the GroupingType for the MetricType
+ * Following algorithm :
```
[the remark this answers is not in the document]
deshsidd: updated

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+        SearchQueryRecord aggregateSearchQueryRecord;
+        String groupId = getGroupingId(searchQueryRecord);
+
+        // New group added to the grouping service
+        // Add to min PQ and overflow records to max PQ (if the number of records in the min PQ exceeds the configured size N)
+        if (!groupIdToAggSearchQueryRecord.containsKey(groupId)) {
+            boolean maxGroupsLimitReached = checkMaxGroupsLimitReached(groupId);
+            if (maxGroupsLimitReached) {
+                return null;
+            }
```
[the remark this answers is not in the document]
deshsidd: `maxGroups` is a setting to limit the number of groups we will process. If we reach the threshold set for max number of query groups we start dropping query groups and do not process them.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Get groupingId. This should be query hashcode for SIMILARITY grouping and user_id for USER_ID grouping.
+     * @param searchQueryRecord record
+     * @return Grouping Id
+     */
+    private String getGroupingId(SearchQueryRecord searchQueryRecord) {
+        switch (groupingType) {
+            case SIMILARITY:
+                return searchQueryRecord.getAttributes().get(Attribute.QUERY_HASHCODE).toString();
```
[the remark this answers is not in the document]
deshsidd: We form the hashcode using the query shape library. This creates a query shape from the query and then hashes the query shape. 2 exact queries with fields shuffled should return the same hashcode.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+import org.opensearch.plugin.insights.rules.model.GroupingType;
+import org.opensearch.plugin.insights.rules.model.MetricType;
+import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
+import org.opensearch.plugin.insights.settings.QueryInsightsSettings;
+
+/**
+ * Handles grouping of search queries based on the GroupingType for the MetricType
+ * Following algorithm :
+ */
+public class QueryGrouper {
```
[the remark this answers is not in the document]
deshsidd: We have one `TopQueriesService` per metricType : latency, cpu and memory. We have all the logic for the exporting query records, rotating window and so on in the TopQueriesService. The grouper is a layer on top of this service which leverages all the above. All it does is group queries by `group_id`. Everything else stays the same as TopQueriesService.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+     * Min heap to keep track of the Top N query groups and is passed from TopQueriesService as the topQueriesStore
+     */
+    private PriorityQueue<SearchQueryRecord> minHeapTopQueriesStore;
+    /**
+     * The Max heap is an overflow data structure used to manage records that exceed the capacity of the Min heap.
+     * It stores all records not included in the Top N query results. When the aggregate measurement for one of these
+     * records is updated and it now qualifies as part of the Top N, the record is moved from the Max heap to the Min heap,
+     * and the records are rearranged accordingly.
+     */
+    private PriorityQueue<SearchQueryRecord> maxHeapQueryStore;
```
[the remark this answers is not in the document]
deshsidd: 1. For a specific query shape the latency, cpu and memory could potentially vary drastically based on the actual query (field used, index being queries and so on) and we cannot assume that all queries with the same shape will exhibit similar resource usage. Keep in mind we are not dealing with queries here but getting the query shape and then hashing it.
2. `maxHeap` limit can be tweaked (cluster setting) based on the requirement. In some edge cases where there are more than 10000 unique queries per window (limit set currently) we could lose out on some queries. But for the majority of the cases we expect the number of unique queries to be less than the threshold. Based on our [benchmarking results](https://github.com/opensearch-project/query-insights/pull/66#issuecomment-2318451568) we see the the number of unique groups are pretty low. Of course this cannot cover all the real world scenarios but we cannot have an unbounded max heap and have to make some reasonable assumptions!

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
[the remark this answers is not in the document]
sgup432: I don't necessarily agree with the current design. My proposal was from a point of view that any store(min/max heap) should reside in a single place so that responsibility of classes are well defined. Currently minHeap store is being shared between TopQueriesService and QueryGrouper which seems wrong as ideally all those things should happen at a single place to ensure thread safety and logical code IMO.

As I see the code, you can easily segregate the logic and move minHeap store to a single place. Plus having different implementation of store also gives a user ability to easily write another implementation and plug that in.

OR other option is to have QueryGrouper only do some grouping logic(as its name suggests) and not perform any get/put operations on the queues. And keep rest of the stuff in service itself.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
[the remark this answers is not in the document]
sgup432: I suggested as groupingType can be updated by thread1 and accessed by thread2 around same time. So to ensure we see the correct groupingType, it is essential to have this volatile.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+        MetricType metricType,
+        GroupingType groupingType,
+        AggregationType aggregationType,
+        PriorityBlockingQueue<SearchQueryRecord> topQueriesStore,
+        int topNSize
+    ) {
+        this.groupingType = groupingType;
+        this.metricType = metricType;
+        this.aggregationType = aggregationType;
+        this.groupIdToAggSearchQueryRecord = new HashMap<>();
```
sgup432, line 94: We also need to make this as concurrentHashMap.

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
```diff
@@ -0,0 +1,292 @@
...
+        this.minHeapTopQueriesStore = topQueriesStore;
+        this.maxHeapQueryStore = new PriorityQueue<>((a, b) -> SearchQueryRecord.compare(b, a, metricType));
```
sgup432, line 96: Need to change this to PriorityBlockingQueue as well

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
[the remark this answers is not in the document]
sgup432: >If only query metrics is enabled we always add the records.
If only top N is enabled we perform an optimization and skip adding the records if they do not make it to the Top N.
If grouping is enabled we need to add all the records since we cannot perform the optimization above.

>Not sure what you are referring to here?

My concern was mostly around the fact that as a user if I disable this framework via setting, we are still adding the records in the background(which are never going to be used) due to grouping being enabled which doesn't look very right to me. I would think less about the optimization/accuracy here compared to correctness.

deshsidd:
As discussed added interfaces and implementation for the following:
interface -> implementation
```
QueryGrouper -> MinMaxHeapQueryGrouper
TopQueriesStore -> PriorityQueueTopQueriesStore
```

src/main/java/org/opensearch/plugin/insights/core/service/QueryGrouper.java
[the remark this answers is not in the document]
deshsidd: As discussed added the following : https://github.com/opensearch-project/query-insights/pull/66#issuecomment-2322535233

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -66,7 +69,7 @@ public class TopQueriesService {
     /**
      * The internal thread-safe store that holds the top n queries insight data
      */
-    private final PriorityQueue<SearchQueryRecord> topQueriesStore;
+    private final PriorityQueueTopQueriesStore<SearchQueryRecord> priorityQueueTopQueriesStore;
```
sgup432, line 72: We can use interface now. PriorityQueueTopQueriesStore -> TopQueriesStore

And rename the variable to ```topQueriesStore```

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -93,6 +96,8 @@ public class TopQueriesService {
      */
     private QueryInsightsExporter exporter;
 
+    private MinMaxHeapQueryGrouper minMaxHeapQueryGrouper;
```
sgup432, line 99: Again we should use interface now, MinMaxHeapQueryGrouper -> QueryGrouper

and rename minMaxHeapQueryGrouper -> queryGrouper

src/main/java/org/opensearch/plugin/insights/core/service/grouper/QueryGrouper.java
```diff
@@ -0,0 +1,66 @@
...
+     * Drain the internal grouping. Needs to be performed after every window or if a setting is changed.
+     */
+    void drain();
+
+    /**
+     * Get the min heap queue that holds the top N queries.
+     *
+     * @return the PriorityBlockingQueue containing the top N queries
+     */
+    PriorityQueueTopQueriesStore<SearchQueryRecord> getMinHeapTopQueriesStore();
```
sgup432, line 37: Same as above. Need to code to interfaces!
PriorityQueueTopQueriesStore -> TopQueriesStore

getMinHeapTopQueriesStore ->  getTopQueriesStore

Need to change javadoc as well accordingly.

src/main/java/org/opensearch/plugin/insights/core/service/grouper/QueryGrouper.java
```diff
@@ -0,0 +1,66 @@
...
+ * Interface for grouping search queries based on grouping type for the metric type.
+ */
+public interface QueryGrouper {
+
+    /**
+     * Add query to the group based on the GroupType setting.
+     * @param searchQueryRecord record to be added
+     * @return the aggregate search query record representing the group
+     */
+    SearchQueryRecord addQueryToGroup(SearchQueryRecord searchQueryRecord);
```
sgup432, line 25: [NIT] I guess we can just call it ```add```? As adding queryRecord is self implied within interface

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -0,0 +1,298 @@
...
+     * Example: Average latency. This query record will be used to store the average latency for multiple query records
+     * in this case.
+     * boolean: True if the aggregate record is in the Top N queries priority query (min heap) and False if the aggregate
+     * record is in the Max Heap
+     */
+    private ConcurrentHashMap<String, Tuple<SearchQueryRecord, Boolean>> groupIdToAggSearchQueryRecord;
+    /**
+     * Min heap to keep track of the Top N query groups and is passed from TopQueriesService as the topQueriesStore
+     */
+    private PriorityQueueTopQueriesStore<SearchQueryRecord> minHeapTopQueriesStore;
```
sgup432, line 61: Need to use interface here too.

deshsidd: Missed these in the IDE refactoring, let me update

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -0,0 +1,298 @@
...
+    @Override
+    public void drain() {
+        log.debug("Number of groups for the current window is " + numberOfGroups());
+        groupIdToAggSearchQueryRecord.clear();
+        maxHeapQueryStore.clear();
+        minHeapTopQueriesStore.clear();
+    }
+
+    @Override
+    public PriorityQueueTopQueriesStore<SearchQueryRecord> getMinHeapTopQueriesStore() {
```
sgup432: I think we don't need to expose this?

sgup432 (approved):

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -69,6 +70,11 @@ public class QueryInsightsSettings {
      */
     public static final String PLUGINS_BASE_URI = "/_insights";
 
+    public static final GroupingType DEFAULT_GROUPING_TYPE = GroupingType.NONE;
+    public static final int DEFAULT_MAX_GROUPS = 100;
+
+    public static final int MAX_GROUPS_LIMIT = 10000;
```
jainankitk, line 76: Can we define the GROUPS limit as function of the number of queries being tracked by top N? 5-10x that limit should be pretty reasonable IMO. Also, can we rename `DEFAULT_MAX_GROUPS` to `DEFAULT_GROUPS_LIMIT`?

deshsidd: We currently manage this as a cluster setting that can be adjusted manually, offering flexibility to increase or decrease the limit as needed. Relying solely on a function of TopN may not be ideal. While setting Max N at 100 would result in a maximum of 1000 groups using the 10x multiplier, this may be insufficient for domains with a high volume of unique queries within the given window. Keeping it adjustable allows for better adaptability to different workloads. 

Sure will do the rename.

src/main/java/org/opensearch/plugin/insights/core/service/store/PriorityQueueTopQueriesStore.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.concurrent.PriorityBlockingQueue;
+
+/**
+ * Implementation of {@link TopQueriesStore} that uses a {@link PriorityBlockingQueue}
+ * to manage Top N search query records or groups. This implementation is thread-safe.
+ *
+ * @param <T> The type of records that this store will manage.
+ */
+public class PriorityQueueTopQueriesStore<T> implements TopQueriesStore<T> {
```
jainankitk, line 21: This implementation and the corresponding interface is not helping IMO. The interface is tightly coupled with the Priority Queue implementation and even this implementation is very thin wrapper on top of Priority Queue. Maybe we can remove this for now, think more carefully the interface and default implementation

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
```diff
@@ -0,0 +1,220 @@
...
+        writeNumber(out, number);
+        out.writeInt(count);
+        out.writeString(aggregationType.toString());
+    }
+
+    private void writeNumber(StreamOutput out, Number number) throws IOException {
+        if (number instanceof Long) {
+            out.writeByte((byte) 0); // Type indicator for Long
+            out.writeLong((Long) number);
+        } else if (number instanceof Integer) {
```
jainankitk, line 171: Why do we need to implement the `Writeable` interface? Probably can get rid of this if not really needed

src/main/java/org/opensearch/plugin/insights/rules/model/Measurement.java
```diff
@@ -0,0 +1,220 @@
...
+                return in.readDouble();
+            case 3:
+                return in.readFloat();
+            default:
+                throw new IOException("Unsupported number type indicator: " + typeIndicator);
+        }
+    }
+
+    public static Measurement readFromStream(StreamInput in) throws IOException {
+        Number number = readNumber(in);
```
jainankitk, line 202: Why do we need to implement the `Writeable` interface? Probably can get rid of this if not really needed

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -130,7 +155,11 @@ public XContentBuilder toXContent(final XContentBuilder builder, final ToXConten
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         out.writeLong(timestamp);
-        out.writeMap(measurements, (stream, metricType) -> MetricType.writeTo(out, metricType), StreamOutput::writeGenericValue);
+        out.writeMap(
+            measurements,
+            (stream, metricType) -> MetricType.writeTo(out, metricType),
+            (stream, measurement) -> measurement.writeTo(out)
+        );
```
jainankitk, line 162: This will break during the version between old and new nodes? I am wondering if we even need to implement the `Writeable` interface for `SearchQueryRecord`?

src/main/java/org/opensearch/plugin/insights/core/service/store/PriorityQueueTopQueriesStore.java
sgup432: +1. TopQueriesStore exposes methods similar to a queue so it doesn't make much sense to have this. My idea was to have a topQueriesStore designed in a way where you can use it to store topN queries on disk/remote etc(or different sinks) and can extend/plug it easily in future if needed. So better to remove this altogether for now.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
deshsidd: This is used here : https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueries.java#L36

src/main/java/org/opensearch/plugin/insights/core/service/store/PriorityQueueTopQueriesStore.java
deshsidd: Removed the interface.

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -112,6 +118,26 @@ public class QueryInsightsSettings {
...
+    public static final Setting<String> TOP_N_QUERIES_GROUP_BY = Setting.simpleString(
+        TOP_N_QUERIES_SETTING_PREFIX + ".group_by",
+        DEFAULT_GROUPING_TYPE.getValue(),
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    /**
+     * Define the group_by option for Top N queries to group queries.
+     */
```
jainankitk, line 133: nit: Can we update the javadoc?

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -112,6 +118,26 @@ public class QueryInsightsSettings {
...
+    public static final Setting<Integer> TOP_N_QUERIES_MAX_GROUPS = Setting.intSetting(
+        TOP_N_QUERIES_SETTING_PREFIX + ".max_groups",
+        DEFAULT_GROUPS_LIMIT,
```
jainankitk, line 136: Can we rename this to something else? `max_groups` is misleading.

ansjcy (changes requested):
The security based integ tests are failing for the change: https://github.com/opensearch-project/query-insights/pull/85
Let's double check if we are missing anything for this change on the permission side before merging.

deshsidd:
> The security based integ tests are failing for the change: https://github.com/opensearch-project/query-insights/pull/85
> Let's double check if we are missing anything for this change on the permission side before merging.

Thanks for checking! The [integration test PR](https://github.com/opensearch-project/query-insights/pull/85) build is failing due to grouping settings not found. This PR needs to be merged for the builds to pass there. The security ITs are run in the checks for this PR. Also ran the security ITs locally and they are passing.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
deshsidd: Made the backward compatibility changes as discussed.

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -0,0 +1,298 @@
...
+    private void updateToMinPQ(SearchQueryRecord searchQueryRecord, SearchQueryRecord aggregateSearchQueryRecord, String groupId) {
+        minHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
+        Number measurementToAdd = searchQueryRecord.getMeasurement(metricType);
+        aggregateSearchQueryRecord.addMeasurement(metricType, measurementToAdd);
+
+        if (maxHeapQueryStore.size() > 0) {
+            addToMaxPQPromoteToMinPQ(aggregateSearchQueryRecord, groupId);
+        } else {
+            addToMinPQOverflowToMaxPQ(aggregateSearchQueryRecord, groupId);
+        }
```
ansjcy, line 239: Why do we need to do this `if/else` here? Can't we simply _always_ add to max/min queue and then do the promote/overflow? 
Also the "else" logic looks effectiveness to me. If the execution enters this else, that means we have already removed a record from the min queue, and max queue is also empty. So when we do `addToMinPQOverflowToMaxPQ` we are adding the previous (but updated) record back it to the min queue again and we won't overflow anything to max queue.

ansjcy: The whole logic can be simplied with:
```
public SearchQueryRecord add(SearchQueryRecord searchQueryRecord) {
    if (!groupIdToAggSearchQueryRecord.containsKey(groupId)) {
        boolean maxGroupsLimitReached = checkMaxGroupsLimitReached(groupId);
        if (maxGroupsLimitReached) {
            return null;
        }
        aggregateSearchQueryRecord = searchQueryRecord;
        aggregateSearchQueryRecord.setGroupingId(groupId);
        aggregateSearchQueryRecord.setMeasurementAggregation(metricType, aggregationType);
        addToMinPQ(aggregateSearchQueryRecord, groupId);
    } else {
        aggregateSearchQueryRecord = groupIdToAggSearchQueryRecord.get(groupId).v1();
        boolean isPresentInMinPQ = groupIdToAggSearchQueryRecord.get(groupId).v2();
        if (isPresentInMinPQ) {
            minHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
        } else {
            maxHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
        }
        addAndPromote(searchQueryRecord, aggregateSearchQueryRecord, groupId);
        
    }
    return aggregateSearchQueryRecord;
}
private void addToMinPQ(SearchQueryRecord searchQueryRecord, String groupId) {
    minHeapTopQueriesStore.add(searchQueryRecord);
    groupIdToAggSearchQueryRecord.put(groupId, new Tuple<>(searchQueryRecord, true));
    overflow();
}
private void addAndPromote(SearchQueryRecord searchQueryRecord, SearchQueryRecord aggregateSearchQueryRecord, String groupId) {
    Number measurementToAdd = searchQueryRecord.getMeasurement(metricType);
    aggregateSearchQueryRecord.addMeasurement(metricType, measurementToAdd);
    addToMinPQ(aggregateSearchQueryRecord, groupId);
    if (maxHeapQueryStore.isEmpty()) {
        return;
    } 
    if (SearchQueryRecord.compare(maxHeapQueryStore.peek(), minHeapTopQueriesStore.peak(), metricType) > 0) {
        SearchQueryRecord recordMovedFromMaxToMin = maxHeapQueryStore.poll();
        addToMinPQ(recordMovedFromMaxToMin, recordMovedFromMaxToMin.getGroupingId());
    }
}
private void overflow() {
    if (minHeapTopQueriesStore.size() > topNSize) {
        SearchQueryRecord recordMovedFromMinToMax = minHeapTopQueriesStore.poll();
        maxHeapQueryStore.add(recordMovedFromMinToMax);
        groupIdToAggSearchQueryRecord.put(recordMovedFromMinToMax.getGroupingId(), new Tuple<>(recordMovedFromMinToMax, false));
    }
}
```

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -0,0 +1,298 @@
...
+
+        while (minHeapTopQueriesStore.size() < topNSize && !maxHeapQueryStore.isEmpty()) {
+            SearchQueryRecord recordMovedFromMaxToMin = maxHeapQueryStore.poll();
+            minHeapTopQueriesStore.add(recordMovedFromMaxToMin);
+            groupIdToAggSearchQueryRecord.put(recordMovedFromMaxToMin.getGroupingId(), new Tuple<>(recordMovedFromMaxToMin, true));
+        }
+    }
+
+    private boolean checkMaxGroupsLimitReached(String groupId) {
+        if (maxGroups <= maxHeapQueryStore.size() && minHeapTopQueriesStore.size() >= topNSize) {
```
ansjcy, line 254: We should emit a metric for this as well.

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -69,6 +70,11 @@ public class QueryInsightsSettings {
      */
     public static final String PLUGINS_BASE_URI = "/_insights";
 
+    public static final GroupingType DEFAULT_GROUPING_TYPE = GroupingType.NONE;
+    public static final int DEFAULT_GROUPS_EXCLUDING_TOPN_LIMIT = 100;
+
+    public static final int MAX_GROUPS_EXCLUDING_TOPN_LIMIT = 10000;
```
ansjcy, line 76: How much memory would 10000 records consume based on the benchmark results?

ansjcy:
> IMHO it might be reasonable to having a setting max_groups to set the maximum number of groups and limit the PQ to that number. If we exceed this number we can drop and add debug logs. The max_groups value should have a validation such that it cannot be set beyond 10,000.

The benchmark numbers looks good, I think it's a good idea to confirm the reasonable upper bound for number of groups as well so that we won't consume too much memory.

ansjcy:
Overall it looks good and I'm fine approving it. But I still have some concerns and we need follow-ups to resolve them.
- Since we are not considering indexed pq in this PR, then `removing an element` in max pq becomes O(number of groups) operation - and remember this is O(number of groups) per search request so potentially it could be very bad. So the possible number of groups in the real world matters a lot in this case. We should have a metric emited to track this very important information so we can make decisions on whether to increase or decrease the max number of groups limit.
- If a search query record is around 1kb then 10000 groups means we will consume 10mb memory at most for this feature, which should be fine. but we still need to watch out on the memory consumption here.
- Some refactoring are needed to simplify the logic in the grouper.

ansjcy (approved):

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * Add query to the group based on the GroupType setting.
+     * The grouping of metrics will be stored within the searchQueryRecord.
+     * @param searchQueryRecord record
+     * @return return the search query record that represents the group
+     */
+    @Override
+    public SearchQueryRecord add(SearchQueryRecord searchQueryRecord) {
```
ansjcy, line 106: This grouper can be simplied with:
```
public SearchQueryRecord add(SearchQueryRecord searchQueryRecord) {
    if (!groupIdToAggSearchQueryRecord.containsKey(groupId)) {
        boolean maxGroupsLimitReached = checkMaxGroupsLimitReached(groupId);
        if (maxGroupsLimitReached) {
            return null;
        }
        aggregateSearchQueryRecord = searchQueryRecord;
        aggregateSearchQueryRecord.setGroupingId(groupId);
        aggregateSearchQueryRecord.setMeasurementAggregation(metricType, aggregationType);
        addToMinPQ(aggregateSearchQueryRecord, groupId);
    } else {
        aggregateSearchQueryRecord = groupIdToAggSearchQueryRecord.get(groupId).v1();
        boolean isPresentInMinPQ = groupIdToAggSearchQueryRecord.get(groupId).v2();
        if (isPresentInMinPQ) {
            minHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
        } else {
            maxHeapTopQueriesStore.remove(aggregateSearchQueryRecord);
        }
        addAndPromote(searchQueryRecord, aggregateSearchQueryRecord, groupId);
        
    }
    return aggregateSearchQueryRecord;
}
private void addToMinPQ(SearchQueryRecord searchQueryRecord, String groupId) {
    minHeapTopQueriesStore.add(searchQueryRecord);
    groupIdToAggSearchQueryRecord.put(groupId, new Tuple<>(searchQueryRecord, true));
    overflow();
}
private void addAndPromote(SearchQueryRecord searchQueryRecord, SearchQueryRecord aggregateSearchQueryRecord, String groupId) {
    Number measurementToAdd = searchQueryRecord.getMeasurement(metricType);
    aggregateSearchQueryRecord.addMeasurement(metricType, measurementToAdd);
    addToMinPQ(aggregateSearchQueryRecord, groupId);
    if (maxHeapQueryStore.isEmpty()) {
        return;
    } 
    if (SearchQueryRecord.compare(maxHeapQueryStore.peek(), minHeapTopQueriesStore.peak(), metricType) > 0) {
        SearchQueryRecord recordMovedFromMaxToMin = maxHeapQueryStore.poll();
        addToMinPQ(recordMovedFromMaxToMin, recordMovedFromMaxToMin.getGroupingId());
    }
}
private void overflow() {
    if (minHeapTopQueriesStore.size() > topNSize) {
        SearchQueryRecord recordMovedFromMinToMax = minHeapTopQueriesStore.poll();
        maxHeapQueryStore.add(recordMovedFromMinToMax);
        groupIdToAggSearchQueryRecord.put(recordMovedFromMinToMax.getGroupingId(), new Tuple<>(recordMovedFromMinToMax, false));
    }
}
```

ansjcy:
Overall it looks good and I'm fine approving it. But I still have some concerns and we need follow-ups to resolve them.
- Since we are not considering indexed pq in this PR, then `removing an element` in max pq becomes O(number of groups) operation - and remember this is O(number of groups) per search request so potentially it could be very bad. So the possible number of groups in the real world matters a lot in this case. We should have a metric emited to track this very important information so we can make decisions on whether to increase or decrease the max number of groups limit.
- If a search query record is around 1kb then 10000 groups means we will consume 10mb memory at most for this feature, which should be fine. but we still need to watch out on the memory consumption here for queries with large source.
- The logic in the grouper is too complicated and refactoring is needed to simplify it.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -39,9 +43,21 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
      */
     public SearchQueryRecord(final StreamInput in) throws IOException, ClassCastException {
         this.timestamp = in.readLong();
-        measurements = new HashMap<>();
-        in.readMap(MetricType::readFromStream, StreamInput::readGenericValue)
-            .forEach(((metricType, o) -> measurements.put(metricType, metricType.parseValue(o))));
+        if (in.getVersion().onOrAfter(Version.V_2_17_0)) {
```
ansjcy, line 46: I'm wondering why is this needed? SearchQueryRecord is only used internally and we are not providing any clients that could cause version mismatch.

deshsidd: This is used when getting the top queries : https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueries.java#L36

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
deshsidd: Did not capture this and we did not reach anywhere close to the 10000 limit in the benchmarks. 

> If a search query record is around 1kb then 10000 groups means we will consume 10mb memory at most for this feature, which should be fine. but we still need to watch out on the memory consumption here for queries with large source.
This analysis seems reasonable but we would need to keep watch out for the memory consumption here.

deshsidd:
Thanks @ansjcy.
1. Will add metrics to track the number of groups discarded as a followup
2. Refactored the logic for grouper

ansjcy (approved):

PR #73, opened 2024-08-23
Query instrumentation Minor Enhancements

deshsidd:
### Description
Minor enhancements to query categorization based on local testing.

1. Make sure query type and aggregation type are 2 separate tags/attributes and do not overlap
2. Change the unit type for histograms for latency, cpu and memory

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
@@ -20,7 +20,7 @@ import org.opensearch.telemetry.metrics.tags.Tags;
  */
 public class SearchQueryAggregationCategorizer {
 
-    private static final String AGGREGATION_TYPE_TAG = "agg_type";
+    static final String AGGREGATION_TYPE_TAG = "agg_type";
     private final SearchQueryCounters searchQueryCounters;
 
     /**
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
@@ -20,7 +20,7 @@ import org.opensearch.telemetry.metrics.tags.Tags;
  */
 public class SearchQueryAggregationCategorizer {
 
-    private static final String TYPE_TAG = "type";
+    private static final String AGGREGATION_TYPE_TAG = "agg_type";
     private final SearchQueryCounters searchQueryCounters;
 
     /**
@@ -49,7 +49,7 @@ public class SearchQueryAggregationCategorizer {
     private void incrementCountersRecursively(AggregationBuilder aggregationBuilder, Map<MetricType, Number> measurements) {
         // Increment counters for the current aggregation
         String aggregationType = aggregationBuilder.getType();
-        searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(TYPE_TAG, aggregationType), measurements);
+        searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(AGGREGATION_TYPE_TAG, aggregationType), measurements);
 
         // Recursively process sub-aggregations if any
         Collection<AggregationBuilder> subAggregations = aggregationBuilder.getSubAggregations();
@@ -63,7 +63,7 @@ public class SearchQueryAggregationCategorizer {
         Collection<PipelineAggregationBuilder> pipelineAggregations = aggregationBuilder.getPipelineAggregations();
         for (PipelineAggregationBuilder pipelineAggregation : pipelineAggregations) {
             String pipelineAggregationType = pipelineAggregation.getType();
-            searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(TYPE_TAG, pipelineAggregationType), measurements);
+            searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(AGGREGATION_TYPE_TAG, pipelineAggregationType), measurements);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
```diff
@@ -49,7 +49,7 @@ public void incrementSearchQueryAggregationCounters(
     private void incrementCountersRecursively(AggregationBuilder aggregationBuilder, Map<MetricType, Number> measurements) {
         // Increment counters for the current aggregation
         String aggregationType = aggregationBuilder.getType();
-        searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(TYPE_TAG, aggregationType), measurements);
+        searchQueryCounters.incrementAggCounter(1, Tags.create().addTag(AGGREGATION_TYPE_TAG, aggregationType), measurements);
```
jainankitk, line 52: The aggregation type will still say just range, right? Or we can tag in the dashboard that the "range" is for aggregation instead of query?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
```diff
@@ -23,8 +23,12 @@
  */
 public final class SearchQueryCounters {
     private static final String LEVEL_TAG = "level";
-    private static final String TYPE_TAG = "type";
+    private static final String QUERY_TYPE_TAG = "type";
     private static final String UNIT = "1";
+    private static final String UNIT_MILLIS = "ms";
+    private static final String UNIT_CPU_CYCLES = "ns";
+    private static final String UNIT_BYTES = "bytes";
```
jainankitk, line 30: Right now this is all in memory. None of these metrics are serialized over the wire, right? Else, we need to think about the backward compatibility aspects

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
```diff
@@ -20,7 +20,7 @@
  */
 public class SearchQueryAggregationCategorizer {
 
-    private static final String TYPE_TAG = "type";
+    private static final String AGGREGATION_TYPE_TAG = "agg_type";
```
jainankitk, line 23: nit: can be package private for use in tests?

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/SearchQueryCategorizerTests.java
```diff
@@ -114,7 +114,7 @@ public void testAggregationsQuery() {
         verify(searchQueryCategorizer.getSearchQueryCounters().getAggCounter()).add(valueCaptor.capture(), tagsCaptor.capture());
 
         double actualValue = valueCaptor.getValue();
-        String actualTag = (String) tagsCaptor.getValue().getTagsMap().get("type");
+        String actualTag = (String) tagsCaptor.getValue().getTagsMap().get("agg_type");
```
jainankitk, line 117: nit: Why not use the `AGGREGATION_TYPE_TAG`?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
```diff
@@ -83,17 +87,17 @@ public SearchQueryCounters(MetricsRegistry metricsRegistry) {
...
             "search.query.type.cpu.histogram",
             "Histogram for the cpu per query type",
-            UNIT
+            UNIT_CPU_CYCLES
         );
         this.queryTypeMemoryHistogram = metricsRegistry.createHistogram(
             "search.query.type.memory.histogram",
             "Histogram for the memory per query type",
-            UNIT
+            UNIT_BYTES
```
jainankitk, line 100: I am assuming the units will not be explicitly written for each record!?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
deshsidd: Yup but the attribute/tag will be `type=range` for query and `agg_type=range` for aggregation

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/SearchQueryCategorizerTests.java
deshsidd: Will have to make it public

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
deshsidd: Nope looks like we are not writing it with every record. Just adding it here for our reference or in case we use it in the future

src/test/java/org/opensearch/plugin/insights/core/service/categorizor/SearchQueryCategorizerTests.java
jainankitk: No that is because of the typo in package name. Don't need public if they are in same package!

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryAggregationCategorizer.java
jainankitk: Nope, that is because of the typo!

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
deshsidd: Yup

jainankitk (approved):

PR #74, opened 2024-08-24
Make sure listener is started when query metrics enabled

deshsidd:
### Description
Listener is only started when Top N is enabled. We need to also start the listener when query metrics is enabled.
Also completed some minor refactoring and updating tests.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -118,7 +118,6 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
      * @param searchQueryMetricsEnabled boolean flag
      */
     public void setSearchQueryMetricsEnabled(boolean searchQueryMetricsEnabled) {
-        log.info("Setting query metrics enabled as : " + searchQueryMetricsEnabled);
         boolean oldSearchQueryMetricsEnabled = queryInsightsService.isSearchQueryMetricsFeatureEnabled();
         this.queryInsightsService.enableSearchQueryMetricsFeature(searchQueryMetricsEnabled);
         if (searchQueryMetricsEnabled) {
@@ -137,9 +136,6 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
     private void updateSettingsForSearchQueryMetrics() {
         clusterService.getClusterSettings()
             .addSettingsUpdateConsumer(SEARCH_QUERY_METRICS_ENABLED_SETTING, v -> setSearchQueryMetricsEnabled(v));
-        log.info(
-            "Setting setSearchQueryMetricsEnabled as : " + clusterService.getClusterSettings().get(SEARCH_QUERY_METRICS_ENABLED_SETTING)
-        );
         setSearchQueryMetricsEnabled(clusterService.getClusterSettings().get(SEARCH_QUERY_METRICS_ENABLED_SETTING));
     }
 
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -114,7 +114,7 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
     }
 
     /**
-     * Enable or disable top queries insights collection for {@link MetricType}
+     * Enable or disable top queries insights collection for {@link MetricType}.
      * This function will enable or disable the corresponding listeners
      * and query insights services.
      *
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -53,11 +54,18 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
...
     @Inject
-    public QueryInsightsListener(final ClusterService clusterService, final QueryInsightsService queryInsightsService) {
+    public QueryInsightsListener(
+        final ClusterService clusterService,
+        final QueryInsightsService queryInsightsService,
+        boolean initiallyEnabled
+    ) {
         this.clusterService = clusterService;
         this.queryInsightsService = queryInsightsService;
+        super.setEnabled(initiallyEnabled); // Disable query insights listener and service initially
```
jainankitk, line 67: Use `super(initiallyEnabled)` instead?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -98,18 +108,61 @@ public void setEnableTopQueries(final MetricType metricType, final boolean isCur
...
+        if (searchQueryMetricsEnabled) {
+            if (!oldSearchQueryMetricsEnabled) {
+                checkAndRestartQueryInsights();
+            }
+        } else {
+            if (oldSearchQueryMetricsEnabled) {
+                checkAndStopQueryInsights();
+            }
+        }
+    }
```
jainankitk, line 137: These logics are starting to get complicated. We should make these stop, starts idempotent and not rely on the old state

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -98,18 +108,61 @@ public void setEnableTopQueries(final MetricType metricType, final boolean isCur
...
+            queryInsightsService.stop();
+            queryInsightsService.start();
+        }
+    }
+
+    private void updateSettingsForSearchQueryMetrics() {
+        clusterService.getClusterSettings()
+            .addSettingsUpdateConsumer(SEARCH_QUERY_METRICS_ENABLED_SETTING, v -> setSearchQueryMetricsEnabled(v));
+        setSearchQueryMetricsEnabled(clusterService.getClusterSettings().get(SEARCH_QUERY_METRICS_ENABLED_SETTING));
+    }
```
jainankitk, line 164: We should not add method for this as this logic for adding settings update consumer should only be invoked once and better done inline

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -98,18 +108,61 @@ public void setEnableTopQueries(final MetricType metricType, final boolean isCur
...
+    /**
+     * Restarts query insights service if any feature enabled
+     */
+    public void checkAndRestartQueryInsights() {
+        if (queryInsightsService.isAnyFeatureEnabled()) {
+            super.setEnabled(true);
+            queryInsightsService.stop();
+            queryInsightsService.start();
+        }
+    }
```
jainankitk, line 158: We can keep the original code, just get boolean return from `checkAndStopQueryInsights` and `checkAndRestartQueryInsights`.

`super.setEnabled(queryInsightsService.checkAndRestartQueryInsights())` or `super.setEnabled(queryInsightsService.checkAndStopQueryInsights())`

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: refactored the logic

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: refactored the logic please take a look

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -56,10 +57,34 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
      */
     @Inject
     public QueryInsightsListener(final ClusterService clusterService, final QueryInsightsService queryInsightsService) {
+        super(false); // Disable query insights listener and service initially
         this.clusterService = clusterService;
         this.queryInsightsService = queryInsightsService;
-        // Setting endpoints set up for top n queries, including enabling top n queries, window size and top n size
-        // Expected metricTypes are Latency, CPU and Memory.
+        initialize();
```
jainankitk, line 63: Can be rewritten as `this(clusterService, queryInsightsService, false);`

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -56,10 +57,34 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
...
+        super(initiallyEnabled);
+        this.clusterService = clusterService;
+        this.queryInsightsService = queryInsightsService;
+        initialize();
+    }
+
+    // Common initialization logic
+    private void initialize() {
+        // Setting endpoints set up for top n queries, including enabling top n queries, window size, and top n size
+        // Expected metricTypes are Latency, CPU, and Memory.
```
jainankitk, line 87: As called out earlier, do it inline instead of having separate method

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -56,10 +57,34 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
...
+     * @param queryInsightsService The topQueriesByLatencyService associated with this listener
+     * @param initiallyEnabled Is the listener initially enabled/disabled
+     */
+    public QueryInsightsListener(
+        final ClusterService clusterService,
+        final QueryInsightsService queryInsightsService,
+        boolean initiallyEnabled
+    ) {
+        super(initiallyEnabled);
+        this.clusterService = clusterService;
```
jainankitk, line 79: Why do we need to store the clusterService reference?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -82,31 +107,48 @@ public QueryInsightsListener(final ClusterService clusterService, final QueryIns
...
+        if (anyFeatureEnabled && !super.isEnabled()) {
             super.setEnabled(true);
-            if (isTopNFeaturePreviouslyDisabled) {
-                queryInsightsService.checkAndRestartQueryInsights();
-            }
+            queryInsightsService.stop();
+            queryInsightsService.start();
+        } else if (!anyFeatureEnabled && super.isEnabled()) {
+            super.setEnabled(false);
+            queryInsightsService.stop();
```
jainankitk, line 151: how is the concurrency handled on this stop/start? For example, if 2 features are enabled/disabled in single settings update, how the listeners for respective settings will consume those updated values. Are the listeners executed on cluster service thread?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: Call to this() has to be the first call in the method body but we already called super(). Don't think we can call both super() and this() in the same constructor.

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: Need in `constructSearchQueryRecord` method

src/test/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListenerTests.java
```diff
@@ -116,7 +117,7 @@ public void testOnRequestEnd() throws InterruptedException {
 
         int numberOfShards = 10;
 
-        QueryInsightsListener queryInsightsListener = new QueryInsightsListener(clusterService, queryInsightsService);
+        QueryInsightsListener queryInsightsListener = new QueryInsightsListener(clusterService, queryInsightsService, false);
```
jainankitk, line 120: nit: this change is not needed. Can use the wrapper constructor?

src/test/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListenerTests.java
```diff
@@ -182,7 +183,7 @@ public void testConcurrentOnRequestEnd() throws InterruptedException {
         CountDownLatch countDownLatch = new CountDownLatch(numRequests);
 
         for (int i = 0; i < numRequests; i++) {
-            searchListenersList.add(new QueryInsightsListener(clusterService, queryInsightsService));
+            searchListenersList.add(new QueryInsightsListener(clusterService, queryInsightsService, false));
```
jainankitk, line 186: nit: this change is not needed. Can use the wrapper constructor?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: `AbstractScopedSettings` handles makes sure the `applySettings` method is synchronized

jainankitk (approved):

PR #84, opened 2024-08-30
Top n queries historical queries from local index and time range

### LINKED ISSUE #12: [FEATURE] Top N Queries API should be able to get historical data from index
Currently we can export historical top n queries data into local index, and to view the historical data we have to manually search the index. It would be convenient if we can provide the functionality to specify time ranges in the top n queries API, and the API handles the logic to search in local index. 

We keep the current logic, but it would not be convenient for the user and also make the dashboard logic complex when getting historical top n queries.

LilyCaroline17:
### Description

When the local index exporter is enabled, historical top N queries data is stored there. Previously, obtaining these historical queries required manually searching through each index. To simplify this process for users, we've introduced two parameters: 'from' and 'to', to the top n queries request. This allows users to define the time range for retrieving historical data. Please ensure that from and to are in ISO format. 

For example, you could make this following call to retrieve the top N queries from August 25, 2024, at 15:00 UTC to August 30, 2024, at 17:00 UTC.
`curl -X GET "http://localhost:9200/_insights/top_queries?from=2024-08-25T15:00:00.000Z&to=2024-08-30T17:00:00.000Z"`

### Issues Resolved
Closes #12 

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
@@ -1,0 +1,127 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.plugin.insights.core.reader;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormatter;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.client.Client;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.index.IndexNotFoundException;
+import org.opensearch.index.query.*;
+import org.opensearch.index.search.MatchQuery;
+import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
+import org.opensearch.search.SearchHit;
+import org.opensearch.search.builder.SearchSourceBuilder;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Local index reader for reading query insights data from local OpenSearch indices.
+ */
+public final class LocalIndexReader implements QueryInsightsReader {
+    /**
+     * Logger of the local index reader
+     */
+    private final Logger logger = LogManager.getLogger();
+    private final Client client;
+    private DateTimeFormatter indexPattern;
+    private final NamedXContentRegistry namedXContentRegistry;
+
+    /**
+     * Constructor of LocalIndexReader
+     *
+     * @param client OS client
+     * @param indexPattern the pattern of index to read from
+     * @param namedXContentRegistry for parsing purposes
+     */
+    public LocalIndexReader(final Client client, final DateTimeFormatter indexPattern, final NamedXContentRegistry namedXContentRegistry) {
+        this.indexPattern = indexPattern;
+        this.client = client;
+        this.namedXContentRegistry = namedXContentRegistry;
+    }
+
+    /**
+     * Getter of indexPattern
+     *
+     * @return indexPattern
+     */
+    public DateTimeFormatter getIndexPattern() {
+        return indexPattern;
+    }
+
+    /**
+     * Setter of indexPattern
+     *
+     * @param indexPattern index pattern
+     * @return the current LocalIndexReader
+     */
+    public LocalIndexReader setIndexPattern(DateTimeFormatter indexPattern) {
+        this.indexPattern = indexPattern;
+        return this;
+    }
+
+    /**
+     * Export a list of SearchQueryRecord from local index
+     *
+     * @param from start timestamp
+     * @param to end timestamp
+     * @return list of SearchQueryRecords whose timestamps fall between from and to
+     */
+    @Override
+    public List<SearchQueryRecord> read(final String from, final String to) {
+        List<SearchQueryRecord> records = new ArrayList<>();
+        final DateTime start = new DateTime(Long.valueOf(from), DateTimeZone.UTC);
+        DateTime end = new DateTime(Long.valueOf(to), DateTimeZone.UTC);
+        if (end.compareTo(DateTime.now()) > 0){
+            end = DateTime.now();
+        }
+        DateTime curr = start;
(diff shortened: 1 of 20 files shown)


deshsidd:
Looks like one of the build is failing please take a look

src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
```diff
@@ -0,0 +1,130 @@
...
+        if (from == null || to == null) {
+            return records;
+        }
+        final DateTime start = DateTime.parse(from);
+        DateTime end = DateTime.parse(to);
+        if (end.compareTo(DateTime.now(DateTimeZone.UTC)) > 0) {
+            end = DateTime.now(DateTimeZone.UTC);
+        }
+        DateTime curr = start;
+        while (curr.compareTo(end.plusDays(1).withTimeAtStartOfDay()) < 0) {
```
ansjcy, line 95: Note: We are making assumptions on the index pattern that it will rollover every day. We should add this assumption with a check in the configuration API as well.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -283,14 +299,16 @@ public void setTopNSize(final MetricType type, final int topNSize) {
...
     /**
-     * Set the exporter config for a metricType
+     * Set the exporter and reader config for a metricType
      *
      * @param type {@link MetricType}
-     * @param settings exporter settings
+     * @param settings exporter and reader settings
      */
-    public void setExporter(final MetricType type, final Settings settings) {
+    public void setExporterReader(final MetricType type, final Settings settings) {
```
ansjcy, line 307: nit: maybe rename to 
```
setExporterAndReader
```
?

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -303,14 +321,15 @@ public SearchQueryCategorizer getSearchQueryCategorizer() {
...
     /**
-     * Validate the exporter config for a metricType
+     * Validate the exporter and reader config for a metricType
      *
      * @param type {@link MetricType}
-     * @param settings exporter settings
+     * @param settings exporter and reader settings
      */
-    public void validateExporterConfig(final MetricType type, final Settings settings) {
+    public void validateExporterReaderConfig(final MetricType type, final Settings settings) {
```
ansjcy, line 329: nit: Same as above, maybe 
```
validateExporterAndReaderConfig
```

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -238,25 +255,62 @@ public void setExporter(final Settings settings) {
...
+     */
+    public void validateExporterReaderConfig(Settings settings) {
         queryInsightsExporterFactory.validateExporterConfig(settings);
+        queryInsightsReaderFactory.validateReaderConfig(settings);
     }
 
+    /**
+     * Lambda function to mark if a record is internal
+     */
+    private final Predicate<SearchQueryRecord> removeInternal = (record) -> {
```
ansjcy, line 284: This lambda function is more about "check if is internal", maybe let's rename it to 
```
checkIfInternal
```
?

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -238,25 +255,62 @@ public void setExporter(final Settings settings) {
...
+        Map<Attribute, Object> attributes = record.getAttributes();
+        Object indicesObject = attributes.get(Attribute.INDICES);
+        if (indicesObject instanceof Object[]) {
+            Object[] indices = (Object[]) indicesObject;
+            return Arrays.stream(indices).noneMatch(index -> {
+                if (index instanceof String) {
+                    String indexString = (String) index;
+                    return indexString.contains("top_queries");
```
ansjcy, line 292: Also we are basically saying ignoring any requests made on the top_queries index. it might be worth discussing it further whether we want to do this.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -267,6 +321,51 @@ public List<SearchQueryRecord> getTopQueriesRecords(final boolean includeLastWin
         if (includeLastWindow) {
             queries.addAll(topQueriesHistorySnapshot.get());
         }
+        List<SearchQueryRecord> filterQueries = queries;
+        if (from != null && to != null) {
+            final DateTime start = DateTime.parse(from);
+            final DateTime end = DateTime.parse(to);
+            Predicate<SearchQueryRecord> timeFilter = element -> start.getMillis() <= element.getTimestamp()
+                && element.getTimestamp() <= end.getMillis();
+            filterQueries = queries.stream().filter(removeInternal.and(timeFilter)).collect(Collectors.toList());
```
ansjcy, line 330: We are applying the filter to remove the search requests to internal indices for in-memory records only. Should we also apply the same filter for historical data in index?

src/main/java/org/opensearch/plugin/insights/rules/transport/top_queries/TransportTopQueriesAction.java
```diff
@@ -86,6 +86,16 @@ protected TopQueriesResponse newResponse(
             default:
                 size = clusterService.getClusterSettings().get(QueryInsightsSettings.TOP_N_LATENCY_QUERIES_SIZE);
         }
+        final String[] timeRange = topQueriesRequest.getTimeRange();
+        if (timeRange[0] != null && timeRange[1] != null) {
```
ansjcy, line 90: Let's check timeRange length==2 before this, otherwise there are possibilities this will cause unexpected exceptions in the future

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
LilyCaroline17: I found that the internal records didn't seem to be stored in the historical data, but I can put in the filter as well just in case.

src/main/java/org/opensearch/plugin/insights/rules/transport/top_queries/TransportTopQueriesAction.java
LilyCaroline17: I'll leave it as checking the length == 2 for now, but feel free to change the getting of from and to!

ansjcy (approved):

ansjcy:
Thanks for making the changes! 
Overall it looks good to me, but we still need to follow up one several things before 2.18 release:
- Discuss further on whether we should ignore any requests made on the `top_queries` index. 
- Have more strict restriction on index pattern the user can use.

Issue #109, opened 2024-09-09
Remove unnecessary map of QueryBuilders to fieldName function


jainankitk:
With the introduction of `WithFieldName` interface in core Opensearch, `QueryBuilder` can directly be checked as `instanceof WithFieldName` instead of maintaining the map of QueryBuilders to fieldName function.

Issue #69, opened 2024-08-07
[FEATURE] [RFC] Query Shape Field Data Type


dzane17:
### Background
Query shape currently has the ability to append **field names**. In addition, we'd like the option to add **field data type** (often called `fieldType` in [code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/query/WildcardQueryBuilder.java#L249)).

In this example, `field1, field2, field3` is are **field names** and `text, date, keyword` are  **data types**.
```
bool
  filter:
    terms [field1:text]
  must:
    range [field2:date]
    term [field3:keyword]
  must_not:
    exists [field4:keyword]
  should:
    match [field5:text]
sort:
  asc [field6:number]
  desc [field2:date]
aggregation:
  terms [field3:keyword]
    aggregation:
      avg [field6:number]
      cardinality [field3:keyword]
      date_histogram [field2:date]
      max [field6:number]
      min [field6:number]
      percentile_ranks [field6:number]
      sum [field6:number]
```

### Problem
It is impossible to determine data type from just the search source.

In this match query, field name is "title", however the **exact** data type is unknown. It could be `text` or `keyword`.
```
GET /my_index/_search
{
  "query": {
    "match_phrase": {
      "title": "Second Document"
    }
  }
}
```

### What solution would you like?
Three possible solutions:
1. Record `dataType` in each *QueryBuilder, *AggregationBuilder, *SortBuilder class during core search execution. Then when records are processed in query-insights plugin we can simply call `aggBuilder.getDataType()`. I have seen builder classes get the data type mapping from shardContext like:
```
final MappedFieldType fieldType = shardContext.fieldMapper(fieldName);
```
Cons: Not all builders look up data type, Need to edit *Builder classes in core

2. Fetch [mappings](https://opensearch.org/docs/latest/field-types/#get-a-mapping) from query-insights plugin. Then we can get data type from known field name.
Note: Need to find a way to fetch _mapping data from query-insights

3. Ignore field data type in query shape

In many cases, data type is known given the query/agg/sort type. For example, data type for date histogram aggregations is always `date`, boolean query is always boolean data type. In these cases, adding data type adds no value:
```
    aggregation:
      date_histogram [date]

vs. 


    aggregation:
      date_histogram
```

On the other hand, Range queries support a variety of data types so we would lose information with this option.
a. Date
```
GET /my_index/_search
{
  "query": {
    "range": {
      "publish_date": {
[...]

PR #111, opened 2024-09-09
Refactor query shape field data maps

dzane17:
### Description
Refactor query shape field data maps

With the introduction of `WithFieldName` interface in OpenSearch core (see https://github.com/opensearch-project/OpenSearch/pull/15705), we can now simplify the field data logic in Query-Insights.

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/109
Related RFC https://github.com/opensearch-project/query-insights/issues/69

### Testing
Large Search
```
curl -XGET 'http://localhost:9200/_search?pretty' -H 'Content-Type: application/json' -d '{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "order_date": {
              "gte": "2023-01-01",
              "lte": "2023-12-31"
            }
          }
        },
        {
          "term": {
            "status": "completed"
          }
        }
      ],
      "filter": [
        {
          "terms": {
            "category": ["electronics", "clothing"]
          }
        }
      ]
    }
  },
  "aggs": {
    "sales_per_month": {
      "date_histogram": {
        "field": "order_date",
        "calendar_interval": "month",
        "format": "yyyy-MM",
        "order": {
          "_key": "desc"
        }
      },
      "aggs": {
        "total_sales": {
          "sum": {
            "field": "total_amount"
          }
        },
        "avg_order_quantity": {
          "avg": {
            "field": "order_quantity"
          }
        },
        "max_order_amount": {
          "max": {
            "field": "total_amount"
          }
        }
      }
    },
    "avg_monthly_sales": {
      "avg_bucket": {
        "buckets_path": "sales_per_month>total_sales"
      }
    },
    "avg_max_order_amount": {
      "avg_bucket": {
        "buckets_path": "sales_per_month>max_order_amount"
      }
    }
  },
  "sort": [
    { "sales_per_month>key": { "order": "desc" } },
    { "total_sales": { "order": "desc" } },
    { "avg_order_quantity": { "order": "asc" } }
  ]
}'
```
```
[2024-09-09T13:15:16,352][INFO ][stdout                   ] [integTest-0] bool []
[2024-09-09T13:15:16,354][INFO ][stdout                   ] [integTest-0]   must:
[2024-09-09T13:15:16,354][INFO ][stdout                   ] [integTest-0]     range [order_date]
[2024-09-09T13:15:16,355][INFO ][stdout                   ] [integTest-0]     term [status]
[2024-09-09T13:15:16,355][INFO ][stdout                   ] [integTest-0]   filter:
[2024-09-09T13:15:16,355][INFO ][stdout                   ] [integTest-0]     terms [category]
[2024-09-09T13:15:16,355][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-09-09T13:15:16,356][INFO ][stdout                   ] [integTest-0]   date_histogram [order_date]
[2024-09-09T13:15:16,356][INFO ][stdout                   ] [integTest-0]     aggregation:
[2024-09-09T13:15:16,356][INFO ][stdout                   ] [integTest-0]       avg [order_quantity]
[2024-09-09T13:15:16,357][INFO ][stdout                   ] [integTest-0]       max [total_amount]
[2024-09-09T13:15:16,357][INFO ][stdout                   ] [integTest-0]       sum [total_amount]
[2024-09-09T13:15:16,357][INFO ][stdout                   ] [integTest-0]   pipeline aggregation:
[2024-09-09T13:15:16,357][INFO ][stdout                   ] [integTest-0]     avg_bucket
[2024-09-09T13:15:16,358][INFO ][stdout                   ] [integTest-0]     avg_bucket
[2024-09-09T13:15:16,358][INFO ][stdout                   ] [integTest-0] sort:
[2024-09-09T13:15:16,358][INFO ][stdout                   ] [integTest-0]   desc [sales_per_month>key]
[2024-09-09T13:15:16,358][INFO ][stdout                   ] [integTest-0]   desc [total_sales]
[2024-09-09T13:15:16,358][INFO ][stdout                   ] [integTest-0]   asc [avg_order_quantity]
```

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
@@ -12,60 +12,15 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.function.Function;
 import org.apache.lucene.util.BytesRef;
 import org.opensearch.common.hash.MurmurHash3;
 import org.opensearch.core.common.io.stream.NamedWriteable;
-import org.opensearch.index.query.AbstractGeometryQueryBuilder;
-import org.opensearch.index.query.CommonTermsQueryBuilder;
-import org.opensearch.index.query.ExistsQueryBuilder;
-import org.opensearch.index.query.FieldMaskingSpanQueryBuilder;
-import org.opensearch.index.query.FuzzyQueryBuilder;
-import org.opensearch.index.query.GeoBoundingBoxQueryBuilder;
-import org.opensearch.index.query.GeoDistanceQueryBuilder;
-import org.opensearch.index.query.GeoPolygonQueryBuilder;
-import org.opensearch.index.query.MatchBoolPrefixQueryBuilder;
-import org.opensearch.index.query.MatchPhrasePrefixQueryBuilder;
-import org.opensearch.index.query.MatchPhraseQueryBuilder;
-import org.opensearch.index.query.MatchQueryBuilder;
-import org.opensearch.index.query.MultiTermQueryBuilder;
-import org.opensearch.index.query.PrefixQueryBuilder;
 import org.opensearch.index.query.QueryBuilder;
-import org.opensearch.index.query.RangeQueryBuilder;
-import org.opensearch.index.query.RegexpQueryBuilder;
-import org.opensearch.index.query.SpanNearQueryBuilder;
-import org.opensearch.index.query.SpanTermQueryBuilder;
-import org.opensearch.index.query.TermQueryBuilder;
-import org.opensearch.index.query.TermsQueryBuilder;
-import org.opensearch.index.query.WildcardQueryBuilder;
+import org.opensearch.index.query.WithFieldName;
 import org.opensearch.search.aggregations.AggregationBuilder;
 import org.opensearch.search.aggregations.AggregatorFactories;
 import org.opensearch.search.aggregations.PipelineAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.histogram.AutoDateHistogramAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.histogram.DateHistogramAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.histogram.HistogramAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.histogram.VariableWidthHistogramAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.missing.MissingAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.range.AbstractRangeBuilder;
-import org.opensearch.search.aggregations.bucket.range.GeoDistanceAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.range.IpRangeAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.sampler.DiversifiedAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.terms.RareTermsAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.terms.SignificantTermsAggregationBuilder;
-import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.AvgAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.CardinalityAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.ExtendedStatsAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.GeoCentroidAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.MaxAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.MinAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.StatsAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.SumAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.ValueCountAggregationBuilder;
-import org.opensearch.search.aggregations.support.ValuesSourceAggregationBuilder;
 import org.opensearch.search.builder.SearchSourceBuilder;
-import org.opensearch.search.sort.FieldSortBuilder;
 import org.opensearch.search.sort.SortBuilder;
 
 /**
@@ -74,9 +29,6 @@ import org.opensearch.search.sort.SortBuilder;
 public class QueryShapeGenerator {
     static final String EMPTY_STRING = "";
     static final String ONE_SPACE_INDENT = " ";
-    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = FieldDataMapHelper.getQueryFieldDataMap();
-    static final Map<Class<?>, List<Function<Object, String>>> AGG_FIELD_DATA_MAP = FieldDataMapHelper.getAggFieldDataMap();
-    static final Map<Class<?>, List<Function<Object, String>>> SORT_FIELD_DATA_MAP = FieldDataMapHelper.getSortFieldDataMap();
 
     /**
      * Method to get query shape hash code given a source
@@ -161,7 +113,7 @@ public class QueryShapeGenerator {
             StringBuilder stringBuilder = new StringBuilder();
             stringBuilder.append(baseIndent).append(ONE_SPACE_INDENT.repeat(2)).append(aggBuilder.getType());
             if (showFields) {
-                stringBuilder.append(buildFieldDataString(AGG_FIELD_DATA_MAP.get(aggBuilder.getClass()), aggBuilder));
+                stringBuilder.append(buildFieldDataString(aggBuilder));
             }
             stringBuilder.append("\n");
 
@@ -227,7 +179,7 @@ public class QueryShapeGenerator {
             StringBuilder stringBuilder = new StringBuilder();
             stringBuilder.append(ONE_SPACE_INDENT.repeat(2)).append(sortBuilder.order());
             if (showFields) {
-                stringBuilder.append(buildFieldDataString(SORT_FIELD_DATA_MAP.get(sortBuilder.getClass()), sortBuilder));
+                stringBuilder.append(buildFieldDataString(sortBuilder));
             }
             shapeStrings.add(stringBuilder.toString());
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -74,9 +34,9 @@
 public class QueryShapeGenerator {
     static final String EMPTY_STRING = "";
     static final String ONE_SPACE_INDENT = " ";
-    static final Map<Class<?>, List<Function<Object, String>>> QUERY_FIELD_DATA_MAP = FieldDataMapHelper.getQueryFieldDataMap();
-    static final Map<Class<?>, List<Function<Object, String>>> AGG_FIELD_DATA_MAP = FieldDataMapHelper.getAggFieldDataMap();
-    static final Map<Class<?>, List<Function<Object, String>>> SORT_FIELD_DATA_MAP = FieldDataMapHelper.getSortFieldDataMap();
+    static final ConcurrentHashMap<Class<?>, List<Method>> QUERY_FIELD_DATA_MAP = FieldDataMapHelper.getQueryFieldDataMap();
+    static final ConcurrentHashMap<Class<?>, List<Method>> AGG_FIELD_DATA_MAP = FieldDataMapHelper.getAggFieldDataMap();
+    static final ConcurrentHashMap<Class<?>, List<Method>> SORT_FIELD_DATA_MAP = FieldDataMapHelper.getSortFieldDataMap();
```
jainankitk, line 39: Do we really need `ConcurrentHashMap`? Also, can we not eliminate the map altogether and check for `isinstanceof WithFieldName` ?

dzane17: Using `ConcurrentHashMap` to eliminate need to check for null keys/values. 

We can use a simple `if` check today, but the maps allows for better extensibility. Assuming we still have other field data in the potential roadmap, I think we can keep the maps.

jainankitk: IMO, having this static Map is not really adding any value. Why do you think the map approach is more extensible?

ansjcy: Is `ConcurrentHashMap` only used to prevent null keys? IMHO if there's no concurrent access cases it might not be worth doing it.

deshsidd: Agreed. Lets skip `ConcurrentHashMap` if this is the case

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -243,11 +203,20 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
-        if (methods != null) {
-            for (Function<Object, String> lambda : methods) {
-                fieldDataList.add(lambda.apply(builder));
+        for (ConcurrentHashMap.Entry<Class<?>, List<Method>> entry : fieldMethodsMap.entrySet()) {
+            if (entry.getKey().isInstance(builder)) {
+                List<Method> methodsList = entry.getValue();
+                for (Method method : methodsList) {
+                    try {
+                        fieldDataList.add((String) method.invoke(builder));
+                    } catch (Exception e) {
```
deshsidd, line 214: nit: can catch the specific exceptions here rather than catchall. Eg:
`catch (IllegalAccessException | InvocationTargetException e) {`

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -257,85 +226,54 @@ static String buildFieldDataString(List<Function<Object, String>> methods, Named
...
-                createEntry(PrefixQueryBuilder.class, PrefixQueryBuilder::fieldName),
-                createEntry(RangeQueryBuilder.class, RangeQueryBuilder::fieldName),
-                createEntry(RegexpQueryBuilder.class, RegexpQueryBuilder::fieldName),
-                createEntry(SpanNearQueryBuilder.SpanGapQueryBuilder.class, SpanNearQueryBuilder.SpanGapQueryBuilder::fieldName),
-                createEntry(SpanTermQueryBuilder.class, SpanTermQueryBuilder::fieldName),
-                createEntry(TermQueryBuilder.class, TermQueryBuilder::fieldName),
-                createEntry(TermsQueryBuilder.class, TermsQueryBuilder::fieldName),
-                createEntry(WildcardQueryBuilder.class, WildcardQueryBuilder::fieldName)
-            );
+        private static ConcurrentHashMap<Class<?>, List<Method>> getQueryFieldDataMap() {
```
deshsidd, line 237: getQueryFieldDataMap, getAggFieldDataMap, and getSortFieldDataMap all follow a similar pattern. Can refactor the logic for the following 3 methods. 

Something similar to:
```
private static <T> ConcurrentHashMap<Class<? extends T>, List<Method>> createFieldDataMap(Class<? extends T> clazz, String methodName) {
    ConcurrentHashMap<Class<? extends T>, List<Method>> map = new ConcurrentHashMap<>();
    try {
        map.put(clazz, List.of(clazz.getMethod(methodName)));
    } catch (NoSuchMethodException e) {
        throw new RuntimeException(INVALID_METHOD_EXCEPTION_STRING, e);
    }
    return map;
}

private static ConcurrentHashMap<Class<?>, List<Method>> getQueryFieldDataMap() {
    return createFieldDataMap(WithFieldName.class, "fieldName");
}

private static ConcurrentHashMap<Class<?>, List<Method>> getAggFieldDataMap() {
    return createFieldDataMap(ValuesSourceAggregationBuilder.class, "field");
}

private static ConcurrentHashMap<Class<?>, List<Method>> getSortFieldDataMap() {
    return createFieldDataMap(FieldSortBuilder.class, "getFieldName");
}
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: @jainankitk Every time we add a new type of builder data, it can potentially be another `if builder instanceof X ...` check.

@ansjcy @deshsidd These maps are read only. I believe ConcurrentHashMap and HashMap are similar for reads as synchronization is only on writes.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Fixed

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: > @ansjcy @deshsidd These maps are read only. I believe ConcurrentHashMap and HashMap are similar for reads as synchronization is only on writes.

Yeah, I am not concerned about `ConcurrentHashMap` since we are writing to this `Map` only during the class initialization. Although, I am questioning if really need to keep it? The `instanceof` check is similar to the `map.contains` check?

dzane17:
Unrelated build failure fix: https://github.com/opensearch-project/query-insights/pull/117

dzane17:
@jainankitk @deshsidd @ansjcy Please take a look again. Since all relevant builders now implement the `WithFieldName` interface, I have combined all three query, aggregation, sort maps into one map.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -243,18 +198,20 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
      * @return String: comma separated list with leading space in square brackets
      * Ex: " [my_field, width:5]"
      */
-    static String buildFieldDataString(List<Function<Object, String>> methods, NamedWriteable builder) {
+    static String buildFieldDataString(Map<Class<?>, List<Function<Object, String>>> fieldMethodsMap, NamedWriteable builder) {
```
jainankitk, line 201: `fieldMethodsMap` paramater can be removed now since we have the same instance across query, sort and aggregation invocations

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -264,78 +221,13 @@ private static <T> Map.Entry<Class<?>, List<Function<Object, String>>> createEnt
...
-        /**
-         * Returns a map where the keys are sort builders, and the values are lists of
-         * functions that extract field values from instances of these classes.
-         *
-         * @return a map with class types as keys and lists of field extraction functions as values.
-         */
-        private static Map<Class<?>, List<Function<Object, String>>> getSortFieldDataMap() {
-            return Map.ofEntries(createEntry(FieldSortBuilder.class, FieldSortBuilder::getFieldName));
+        private static Map<Class<?>, List<Function<Object, String>>> getFieldDataMethodsMap() {
+            return Map.ofEntries(createEntry(WithFieldName.class, WithFieldName::fieldName));
```
jainankitk, line 230: given we have single instance today, no need to have this unnecessary map. We can include it if needed later

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -243,18 +198,20 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
      * Ex: " [my_field, width:5]"
      */
-    static String buildFieldDataString(List<Function<Object, String>> methods, NamedWriteable builder) {
+    static String buildFieldDataString(Map<Class<?>, List<Function<Object, String>>> fieldMethodsMap, NamedWriteable builder) {
         List<String> fieldDataList = new ArrayList<>();
-        if (methods != null) {
-            for (Function<Object, String> lambda : methods) {
-                fieldDataList.add(lambda.apply(builder));
+        for (Map.Entry<Class<?>, List<Function<Object, String>>> entry : fieldMethodsMap.entrySet()) {
+            if (entry.getKey().isInstance(builder)) {
```
jainankitk, line 204: Let us just use `builder instanceof WithFieldName` for now?

jainankitk (approved):

PR #113, opened 2024-09-10
Use separate targets for security enabled tests

ansjcy:
### Description
Use separate targets for security enabled tests. 


deshsidd (approved):

deshsidd:
Thanks, is this the reason why the ITs are failing?

ansjcy:
@deshsidd no, we need to look into the failures later, but it should be safe to disable them since we already have a dedicated target to run integration tests with security.

Issue #9, opened 2024-06-18
[FEATURE] Enrich the Operational Metrics in Query Insights


ansjcy:
### Is your feature request related to a problem?
Currently, we have error logs for Query Insights, but we need to introduce more operational metrics so we can create more alarms (e.g., memory usage, queue sizes) for Query Insights features.

### What solution would you like?
Introduce additional metrics, potentially using the OpenTelemetry metric framework. The metrics should comprehensively cover:
- Size information for top queries priority queues
- Size information for multiple blocking queues
- Exported data size
- Errors and exceptions

dblock:
[Catch All Triage, attendees [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/lukas-vlcek/), [4](https://github.com/rafaelma), [5](https://github.com/krisfreedain), [6](https://github.com/gargrajat), [7](https://github.com/andrross)]

ansjcy:
We need to pick this up for 2.18.

ansjcy:
We want to provide a `health_stats` API to expose the important metrics in query insights like queue sizes. Also, we need to emit metrics when error happens (possibly through oTel).

PR #120, opened 2024-09-17
Add data models for health stats API

ansjcy:
### Description
Add data models and related unit tests for health stats API. It includes health stats data model for:
- top queries service
- query grouper service
- query insights service

### Issues Resolved
Related to https://github.com/opensearch-project/query-insights/issues/9

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryGrouperHealthStats.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryGrouperHealthStats.java
@@ -1,0 +1,92 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.plugin.insights.rules.model.healthStats;
+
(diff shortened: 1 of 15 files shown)


src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryGrouperHealthStats.java
```diff
@@ -0,0 +1,92 @@
...
+import org.opensearch.core.xcontent.XContentBuilder;
+
+/**
+ * Represents the health statistics of the query grouper.
+ */
+public class QueryGrouperHealthStats implements ToXContentFragment, Writeable {
+    private final int queryGroupCount;
+    private final int queryGroupHeapSize;
+    private static final String QUERY_GROUP_COUNT = "QueryGroupCount";
+    private static final String QUERY_GROUP_HEAP_SIZE = "QueryGroupHeapSize";
```
dzane17, line 25: I originally thought this measured memory usage. Wondering if we should rename to "count" to avoid confusion even though we have multiple underlying data structures (priority queue, max heap).

ansjcy: Updated!

dzane17 (approved):

PR #130, opened 2024-09-26
Add field type to query shape

dzane17:
### Description
Add field type to query shape

### Issues Resolved
Field type RFC https://github.com/opensearch-project/query-insights/issues/69

### Testing
```
% curl -X POST "localhost:9200/my_index/_search?pretty" -H 'Content-Type: application/json' -d '{
  "query": {
    "range": {
      "age": {
        "gte": 25,
        "lte": 40
      }
    }
  },
  "aggs": {
    "average_age": {
      "avg": {
        "field": "age"
      }
    }
  },
  "sort": [
    {
      "name": {
        "order": "asc"
      }
    }
  ],
  "size": 100
}'
```
```
[2024-09-26T17:07:32,230][INFO ][stdout                   ] [integTest-0] ========== SEARCH SHAPE ==========
[2024-09-26T17:07:32,230][INFO ][stdout                   ] [integTest-0] range [age, integer]
[2024-09-26T17:07:32,231][INFO ][stdout                   ] [integTest-0] aggregation:
[2024-09-26T17:07:32,231][INFO ][stdout                   ] [integTest-0]   avg [age, integer]
[2024-09-26T17:07:32,231][INFO ][stdout                   ] [integTest-0] sort:
[2024-09-26T17:07:32,232][INFO ][stdout                   ] [integTest-0]   asc [name, keyword]
```

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -239,7 +240,16 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
                 );
             }
 
-            String hashcode = QueryShapeGenerator.getShapeHashCodeAsString(request.source(), false);
+            Set<String> successfulShardIndices = searchRequestContext.getSuccessfulSearchShardIndices();
+            if (successfulShardIndices == null) {
+                successfulShardIndices = Collections.emptySet();
+            }
+            String hashcode = QueryShapeGenerator.getShapeHashCodeAsString(
+                request.source(),
+                false,
+                clusterService.state().getMetadata(),
+                successfulShardIndices
+            );
 
             Map<Attribute, Object> attributes = new HashMap<>();
             attributes.put(Attribute.SEARCH_TYPE, request.searchType().toString().toLowerCase(Locale.ROOT));
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
@@ -12,7 +12,12 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import org.apache.lucene.util.BytesRef;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.metadata.MappingMetadata;
+import org.opensearch.cluster.metadata.Metadata;
 import org.opensearch.common.hash.MurmurHash3;
 import org.opensearch.core.common.io.stream.NamedWriteable;
 import org.opensearch.index.query.QueryBuilder;
@@ -34,16 +39,28 @@ public class QueryShapeGenerator {
      * Method to get query shape hash code given a source
      * @param source search request source
      * @param showFields whether to include field data in query shape
+     * @param metadata contains index mappings
+     * @param searchIndices successful indices searched
      * @return Hash code of query shape as a MurmurHash3.Hash128 object (128-bit)
      */
-    public static MurmurHash3.Hash128 getShapeHashCode(SearchSourceBuilder source, Boolean showFields) {
-        final String shape = buildShape(source, showFields);
+    public static MurmurHash3.Hash128 getShapeHashCode(
+        SearchSourceBuilder source,
+        Boolean showFields,
+        Metadata metadata,
+        Set<String> searchIndices
+    ) {
+        final String shape = buildShape(source, showFields, metadata, searchIndices);
         final BytesRef shapeBytes = new BytesRef(shape);
         return MurmurHash3.hash128(shapeBytes.bytes, 0, shapeBytes.length, 0, new MurmurHash3.Hash128());
     }
 
-    public static String getShapeHashCodeAsString(SearchSourceBuilder source, Boolean showFields) {
-        MurmurHash3.Hash128 hashcode = getShapeHashCode(source, showFields);
+    public static String getShapeHashCodeAsString(
+        SearchSourceBuilder source,
+        Boolean showFields,
+        Metadata metadata,
+        Set<String> searchIndices
+    ) {
+        MurmurHash3.Hash128 hashcode = getShapeHashCode(source, showFields, metadata, searchIndices);
         String hashAsString = Long.toHexString(hashcode.h1) + Long.toHexString(hashcode.h2);
         return hashAsString;
     }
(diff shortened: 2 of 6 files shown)


src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -239,7 +240,16 @@ private void constructSearchQueryRecord(final SearchPhaseContext context, final
                 );
             }
 
-            String hashcode = QueryShapeGenerator.getShapeHashCodeAsString(request.source(), false);
+            Set<String> successfulShardIndices = searchRequestContext.getSuccessfulSearchShardIndices();
+            if (successfulShardIndices == null) {
+                successfulShardIndices = Collections.emptySet();
+            }
+            String hashcode = QueryShapeGenerator.getShapeHashCodeAsString(
```
deshsidd, line 247: We should have a knob to enable/disable field type in the query shape?

deshsidd: Or do we plan to show this if fieldName enabled?

deshsidd: Do we want separate knobs for field name and field type?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +225,60 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+        for (String searchIndex : searchIndices) {
+            IndexMetadata indexMetadata = metadata.index(searchIndex);
+            if (indexMetadata == null) {
+                continue;
+            }
+            MappingMetadata mappingMetadata = indexMetadata.mapping();
+            if (mappingMetadata == null) {
+                continue;
+            }
+            Map<String, Object> sourceMap = mappingMetadata.getSourceAsMap();
```
deshsidd, line 258: Can this be null or not of expected format as a corner case?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +225,60 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+
+            String fieldType = findFieldType((Map<String, Object>) sourceMap.get("properties"), fieldName);
+            if (fieldType != null) {
```
deshsidd, line 261: So we are just returning the first field type we find?
I thought we were planning to get field type for all indices and if the field type is the same then return it else consider it ambiguous and process it differently?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +225,60 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+        return null;
+    }
+
+    public static String findFieldType(Map<String, Object> properties, String fieldName) {
+        for (Map.Entry<String, Object> entry : properties.entrySet()) {
+            Map<String, Object> field = (Map<String, Object>) entry.getValue();
+            if (entry.getKey().equals(fieldName)) {
+                return (String) field.get("type");
+            }
+            // TODO: Add support for multi-fields
```
deshsidd, line 274: When can we add this?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
dzane17: I did not think of a good way to handle this. A new method param for each field data is not sustainable. Maybe a list of string constants or enum would work best. 

Adding knobs is not complex. Given that showFields is currently disabled, I vote we implement knobs when we decide to turn on showFields. This way we can tailor it directly for our needs and not spend extra effort now.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: My method is assuming a fieldName will be mapped to the same fieldType in all indices a large majority of the time. Also note, we are only checking the index mapping of indices which were searched as part of this request. 

We do not have any concrete data on this so I am open to either method. The current implementation saves some time and memory (very little).

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: This will need to come in a future iteration. The challenge is that a single multi field can be queried with `name`, `name.keyword`, `name.lowercase`, `name.initials` with the example mapping below. Also non-multi fields can contain "." in the field name. So `name.initials` could be a multi field or just a normal field name.
```
{
  "mappings": {
    "properties": {
      "name": {
        "type": "text",
        "fields": {
          "keyword": {
            "type": "keyword",
            "ignore_above": 256
          },
          "lowercase": {
            "type": "text",
            "normalizer": "lowercase_normalizer"
          },
          "initials": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword",
                "ignore_above": 256
              }
            }
          }
        }
      }
    }
  },
  "settings": {
    "analysis": {
      "normalizer": {
        "lowercase_normalizer": {
          "type": "custom",
          "filter": ["lowercase"]
        }
      }
    }
  }
}
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +225,60 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+            }
+            Map<String, Object> sourceMap = mappingMetadata.getSourceAsMap();
+
+            String fieldType = findFieldType((Map<String, Object>) sourceMap.get("properties"), fieldName);
+            if (fieldType != null) {
+                return fieldType;
+            }
+        }
+        return null;
+    }
```
jainankitk, line 266: Can we inject the `IndicesServices` into QueryInsightsPlugin and do something like below:

```
MappedFieldType mappedFieldType = indicesService.indexService(index).mapperService().fieldType(fieldName)
if (mappedFieldType != null) {
    return mappedFieldType.typeName();
}
```

jainankitk: We might need to update ```results.getSuccessfulResults().map(result -> result.getSearchShardTarget().getIndex()``` to ```results.getSuccessfulResults().map(result -> result.getSearchShardTarget().getShardId().getIndex()``` for getting the `Index` object instead of `String`

dzane17: `QueryInsightsPlugin` is instantiated in core in `node.java` with the [createComponents()](https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java#L66) method. This method `createComponents()` overrides the `Plugin/TelemetryAwarePlugin` class. We will need to overload createComponents to add IndicesServices.

dzane17: TelemetryAwarePlugin is already one example of overloading createComponents() to inject telemetry related classes to relevant plugins. We would need something similar.

PR #135, opened 2024-10-08
Add grouping settings for query field name and type

deshsidd:
### Description

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -9,11 +9,7 @@
 package org.opensearch.plugin.insights.core.listener;
 
 import static org.opensearch.plugin.insights.settings.QueryCategorizationSettings.SEARCH_QUERY_METRICS_ENABLED_SETTING;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_QUERIES_GROUP_BY;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_QUERIES_MAX_GROUPS_EXCLUDING_N;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.getTopNEnabledSetting;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.getTopNSizeSetting;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.getTopNWindowSizeSetting;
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -54,6 +56,8 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
 
     private final QueryInsightsService queryInsightsService;
     private final ClusterService clusterService;
+    private boolean groupingFieldNameEnabled;
+    private boolean groupingFieldTypeEnabled;
```
ansjcy, line 60: Where do we read this? do we need a follow up PR to use it?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -241,7 +262,10 @@ private void constructSearchQueryRecord(final SearchPhaseContext context, final
                 );
             }
 
-            String hashcode = QueryShapeGenerator.getShapeHashCodeAsString(request.source(), false);
+            String hashcode = "";
```
ansjcy, line 265: Let's just don't add `QUERY_HASHCODE` as an attribute if it is not enabled. There's no point to add an empty string to the attribute map.

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -83,6 +83,8 @@ public class QueryInsightsSettings {
     public static final String TOP_QUERIES_BASE_URI = PLUGINS_BASE_URI + "/top_queries";
     /** Default prefix for top N queries feature */
     public static final String TOP_N_QUERIES_SETTING_PREFIX = "search.insights.top_queries";
+    /** Default prefix for top N queries grouping feature */
+    public static final String TOP_N_QUERIES_GROUPING_SETTING_PREFIX = "search.insights.top_queries.grouping";
```
ansjcy, line 87: As I think about it more, I think it still makes sense to combine all the grouping config under the same config group. Let's add a task for this for 2.18 or a future release?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
deshsidd: Yup @dzane17 's PR will use this

dzane17 (approved):

deshsidd:
https://github.com/opensearch-project/query-insights/issues/136 cc @ansjcy

PR #137, opened 2024-10-09
Add query shape field type

dzane17:
### Description
Add query shape field type

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/MappingVisitor.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/MappingVisitor.java
@@ -1,0 +1,76 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
+ * Licensed to Elasticsearch under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+package org.opensearch.plugin.insights.core.service.categorizer;
+
+import java.util.Map;
+
+/**
+ * Visitor pattern for obtaining index mappings
+ *
+ * @opensearch.internal
+ */
+final class MappingVisitor {
+
+    private MappingVisitor() {}
+
+    @SuppressWarnings("unchecked")
+    static String visitMapping(Map<String, ?> propertiesAsMap, String[] fieldName, int idx) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
```diff
@@ -87,11 +87,6 @@ public void categorize(SearchQueryRecord record) {
         incrementQueryTypeCounters(source.query(), measurements);
         incrementQueryAggregationCounters(source.aggregations(), measurements);
         incrementQuerySortCounters(source.sorts(), measurements);
-
-        if (logger.isTraceEnabled()) {
```
deshsidd, line 91: Why did we remove this?

deshsidd:
I am working on the query type changes. Will open a new PR with the changes. Also need to integrate with the settings changes and make sure we can handle field type and field name enabled separately.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/MappingVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+package org.opensearch.plugin.insights.core.service.categorizer;
+
+import java.util.Map;
+
+/**
+ * Visitor pattern for obtaining index mappings
+ *
+ * @opensearch.internal
+ */
+final class MappingVisitor {
```
jainankitk, line 42: This is not really `MappingVisitor`, since we don't want to visit all the mappings, but rather specific field we are interested in

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/MappingVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private MappingVisitor() {}
+
+    @SuppressWarnings("unchecked")
+    static String visitMapping(Map<String, ?> propertiesAsMap, String[] fieldName, int idx) {
+        if (propertiesAsMap.containsKey(fieldName[idx])) {
+            Map<String, ?> fieldMapping = (Map<String, ?>) propertiesAsMap.get(fieldName[idx]);
+            if (idx == fieldName.length - 1) {
+                return (String) fieldMapping.get("type");
```
jainankitk, line 51: Let us not assume the presence of `type` within fieldMapping. This can potentially throw exception

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +216,47 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+                .metadata()
+                .findMappings(successfulSearchShardIndices.stream().map(Index::getName).toArray(String[]::new), input -> (str -> true));
+        } catch (IOException e) {
+            return null;
+        }
+        for (Index index : successfulSearchShardIndices) {
+            MappingMetadata mappingMetadata = allIndicesMap.get(index.getName());
+            if (mappingMetadata != null) {
+                @SuppressWarnings("unchecked")
+                Map<String, ?> propertiesAsMap = (Map<String, ?>) mappingMetadata.getSourceAsMap().get("properties");
```
jainankitk, line 251: Can we preserve this `propertiesAsMap` for an index for the search request? Don't really want to create parse the same mappingMetadata for getting different fieldTypes within same index

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -195,11 +216,47 @@ static String buildSortShape(List<SortBuilder<?>> sortBuilderList, Boolean showF
...
+                String fieldType = MappingVisitor.visitMapping(propertiesAsMap, fieldName.split("\\."), 0);
+                if (fieldType != null) {
+                    // add item to cache
+                    fieldTypeMap.computeIfAbsent(index, k -> new ConcurrentHashMap<>()).put(fieldName, fieldType);
```
jainankitk, line 255: Should we use `computeIfAbsent` instead of `put` even for the inner method?

PR #142, opened 2024-10-10
Adding cache eviction and listener for invalidating index field type …

jainankitk:
…mappings on index deletion/update

### Description
This change primarily achieves 2 things:
* Introduces cache eviction logic for ensuring that the field data cache stays bounded
* Adds cluster state listener for invalidating index field type mappings on index deletion/update

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
@@ -1,3 +1,11 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
 package org.opensearch.plugin.insights.core.service.categorizer;
 
 import java.util.concurrent.ConcurrentHashMap;
@@ -8,23 +16,21 @@ import org.apache.lucene.util.RamUsageEstimator;
 import org.opensearch.common.cache.Cache;
 import org.opensearch.common.cache.CacheBuilder;
 import org.opensearch.common.metrics.CounterMetric;
-import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
-import org.opensearch.core.common.unit.ByteSizeValue;
 import org.opensearch.core.index.Index;
+import org.opensearch.plugin.insights.settings.QueryCategorizationSettings;
 
+/**
+ * Cache implementation specifically for maintaining the field name type mappings
+ * for indices that are part of successful search requests
+ */
 public class IndicesFieldTypeCache {
 
     private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
-    public static final Setting<ByteSizeValue> INDICES_FIELD_TYPE_CACHE_SIZE_KEY = Setting.memorySizeSetting(
-        "search.insights.indices.fieldtype.cache.size",
-        new ByteSizeValue(-1),
-        Setting.Property.NodeScope
-    );
     private final Cache<Index, IndexFieldMap> cache;
 
     public IndicesFieldTypeCache(Settings settings) {
-        final long sizeInBytes = -1; // TODO: INDICES_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
+        final long sizeInBytes = QueryCategorizationSettings.SEARCH_QUERY_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
         CacheBuilder<Index, IndexFieldMap> cacheBuilder = CacheBuilder.<Index, IndexFieldMap>builder();
         if (sizeInBytes > 0) {
             cacheBuilder.setMaximumWeight(sizeInBytes).weigher((k, v) -> RamUsageEstimator.sizeOfObject(k) + v.weight());
@@ -39,6 +45,9 @@ public class IndicesFieldTypeCache {
             logger.error("Unexpected execution exception while initializing for index " + index);
         }
 
+        // Should never return null as the ExecutionException is only thrown
+        // if loader throws an exception or returns a null value, which cannot
+        // be the case in this scenario
         return null;
     }
 
@@ -64,7 +73,7 @@ public class IndicesFieldTypeCache {
         }
 
         public void putIfAbsent(String key, String value) {
-            // Increment the weight only if the key value got added to the Map
+            // Increment the weight only if the key value pair added to the Map
             if (fieldTypeMap.putIfAbsent(key, value) == null) {
                 weight.inc(RamUsageEstimator.sizeOf(key) + RamUsageEstimator.sizeOf(value));
             }
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
@@ -1,0 +1,78 @@
+package org.opensearch.plugin.insights.core.service.categorizer;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.util.RamUsageEstimator;
+import org.opensearch.common.cache.Cache;
+import org.opensearch.common.cache.CacheBuilder;
+import org.opensearch.common.metrics.CounterMetric;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.common.unit.ByteSizeValue;
+import org.opensearch.core.index.Index;
+import org.opensearch.indices.fielddata.cache.IndicesFieldDataCache;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+
+public class IndicesFieldTypeCache {
+
+    private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
+    public static final Setting<ByteSizeValue> INDICES_FIELD_TYPE_CACHE_SIZE_KEY = Setting.memorySizeSetting(
+        "search.insights.indices.fieldtype.cache.size",
+        new ByteSizeValue(-1),
+        Setting.Property.NodeScope
+    );
+    private final Cache<Index, IndexFieldMap> cache;
+
+    public IndicesFieldTypeCache(Settings settings) {
+        final long sizeInBytes = -1; //TODO: INDICES_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
+        CacheBuilder<Index, IndexFieldMap> cacheBuilder = CacheBuilder.<Index, IndexFieldMap>builder();
+        if (sizeInBytes > 0) {
+            cacheBuilder.setMaximumWeight(sizeInBytes).weigher((k, v) -> RamUsageEstimator.sizeOfObject(k) + v.weight());
+        }
+        cache = cacheBuilder.build();
+    }
+
+    public IndexFieldMap getOrInitialize(Index index) {
+        try {
+            return cache.computeIfAbsent(index, k -> new IndexFieldMap());
+        } catch (ExecutionException ex) {
+            logger.error("Unexpected execution exception while initializing for index " + index);
+        }
+
+        return null;
+    }
+
+    public void invalidate(Index index) {
+        cache.invalidate(index);
+    }
+
+    public Iterable<Index> keySet() {
+        return cache.keys();
+    }
+
+    static class IndexFieldMap {
+        private ConcurrentHashMap<String, String> fieldTypeMap;
+        private CounterMetric weight;
+        IndexFieldMap() {
+            fieldTypeMap = new ConcurrentHashMap<>();
+            weight = new CounterMetric();
+        }
+
+        public String get(String fieldName) {
+            return fieldTypeMap.get(fieldName);
+        }
+
+        public void putIfAbsent(String key, String value) {
+            // Increment the weight only if the key value got added to the Map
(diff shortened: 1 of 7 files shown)


deshsidd:
Existing tests still failing. Please take a look @jainankitk

deshsidd:
Looks like one of the reasons of the build failures is the license headers:
```
X:/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+import org.opensearch.common.metrics.CounterMetric;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.common.unit.ByteSizeValue;
+import org.opensearch.core.index.Index;
+
+public class IndicesFieldTypeCache {
+
+    private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
+    public static final Setting<ByteSizeValue> INDICES_FIELD_TYPE_CACHE_SIZE_KEY = Setting.memorySizeSetting(
```
deshsidd, line 19: Might want to add these in one of the `settings` files to keep things consistent

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.util.RamUsageEstimator;
+import org.opensearch.common.cache.Cache;
+import org.opensearch.common.cache.CacheBuilder;
+import org.opensearch.common.metrics.CounterMetric;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.common.unit.ByteSizeValue;
+import org.opensearch.core.index.Index;
+
```
deshsidd, line 15: nit: can add comment for what the class does

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
+    public static final Setting<ByteSizeValue> INDICES_FIELD_TYPE_CACHE_SIZE_KEY = Setting.memorySizeSetting(
+        "search.insights.indices.fieldtype.cache.size",
+        new ByteSizeValue(-1),
+        Setting.Property.NodeScope
+    );
+    private final Cache<Index, IndexFieldMap> cache;
+
+    public IndicesFieldTypeCache(Settings settings) {
+        final long sizeInBytes = -1; // TODO: INDICES_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
```
deshsidd, line 27: Why is this change a TODO can we not make the change now?

ansjcy, line 27: Is `sizeInBytes = -1` intended for this release?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+import org.opensearch.core.common.unit.ByteSizeValue;
+import org.opensearch.core.index.Index;
+
+public class IndicesFieldTypeCache {
+
+    private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
+    public static final Setting<ByteSizeValue> INDICES_FIELD_TYPE_CACHE_SIZE_KEY = Setting.memorySizeSetting(
+        "search.insights.indices.fieldtype.cache.size",
+        new ByteSizeValue(-1),
+        Setting.Property.NodeScope
```
deshsidd, line 22: Will need to add a settings update consumer for these and method to validate and set?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+    );
+    private final Cache<Index, IndexFieldMap> cache;
+
+    public IndicesFieldTypeCache(Settings settings) {
+        final long sizeInBytes = -1; // TODO: INDICES_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
+        CacheBuilder<Index, IndexFieldMap> cacheBuilder = CacheBuilder.<Index, IndexFieldMap>builder();
+        if (sizeInBytes > 0) {
+            cacheBuilder.setMaximumWeight(sizeInBytes).weigher((k, v) -> RamUsageEstimator.sizeOfObject(k) + v.weight());
```
deshsidd, line 30: The RamUsageEstimator.sizeOfObject(k) and v.weight() could be expensive calculations depending on the object size. If this cache handles a lot of entries, calculating memory usage on each insert could have performance implications?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+        cache.invalidate(index);
+    }
+
+    public Iterable<Index> keySet() {
+        return cache.keys();
+    }
+
+    static class IndexFieldMap {
+        private ConcurrentHashMap<String, String> fieldTypeMap;
+        private CounterMetric weight;
```
deshsidd, line 55: Do we need to handle concurrent access for weight?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public String get(String fieldName) {
+            return fieldTypeMap.get(fieldName);
+        }
+
+        public void putIfAbsent(String key, String value) {
+            // Increment the weight only if the key value got added to the Map
+            if (fieldTypeMap.putIfAbsent(key, value) == null) {
+                weight.inc(RamUsageEstimator.sizeOf(key) + RamUsageEstimator.sizeOf(value));
```
deshsidd, line 69: Could this be potentially expensive operations and can they be avoided?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -33,16 +34,36 @@
...
 
     public QueryShapeGenerator(ClusterService clusterService) {
         this.clusterService = clusterService;
-        this.fieldTypeMap = new ConcurrentHashMap<>();
+        clusterService.addListener(this);
+        this.indicesFieldTypeCache = new IndicesFieldTypeCache(clusterService.getSettings());
+    }
+
+    public void clusterChanged(ClusterChangedEvent event) {
+        final List<Index> indicesDeleted = event.indicesDeleted();
```
deshsidd, line 51: Don't think this can be null since we seem to be returning an empty list in the case nothing was deleted but please verify

ansjcy (approved):
LGTM, left 1 minor comment. 
@jainankitk do you think we should add the cache size as part of the healthcheck API as well?
Also we need to add the license header as caught by the build.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -33,16 +34,36 @@
...
+        for (Index index : indicesDeleted) {
+            // remove the deleted index mapping from field type cache
+            indicesFieldTypeCache.invalidate(index);
+        }
+
+        if (event.metadataChanged()) {
+            final Metadata previousMetadata = event.previousState().metadata();
+            final Metadata currentMetadata = event.state().metadata();
+            for (Index index : indicesFieldTypeCache.keySet()) {
```
deshsidd, line 60: We need to check this incase some indices have metadata changed and others not, correct?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -127,20 +148,12 @@ public String buildShape(
...
-        try {
-            indexMapping = clusterService.state().metadata().findMappings(new String[] { index.getName() }, input -> str -> true);
-        } catch (IOException e) {
-            // If an error occurs while retrieving mappings, return an empty map
-            return Collections.emptyMap();
-        }
-
-        MappingMetadata mappingMetadata = indexMapping.get(index.getName());
-        if (mappingMetadata == null) {
+        IndexMetadata indexMetadata = clusterService.state().metadata().index(index);
```
deshsidd, line 151: Did this work with the existing tests?

deshsidd:
> LGTM, left 1 minor comment. @jainankitk do you think we should add the cache size as part of the healthcheck API as well? Also we need to add the license header as caught by the build.

would  be good to have this as part of the health check APIs in case we see performance issues and need to debug

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
ansjcy: `RamUsageEstimator.sizeOf(key)` is O(1)
```
 public static long sizeOf(String s) {
    if (s == null) {
      return 0L;
    } else {
      long size = (long)STRING_SIZE + (long)NUM_BYTES_ARRAY_HEADER + 2L * (long)s.length();
      return alignObjectSize(size);
    }
  }
```

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
jainankitk: Yeah, the weight calculation for each entry in the top level map is O(1) operation. The primary reason we are maintaining weight metric in per index map

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: Yes, iterating over the list of existing indices in the cache to invalidate the ones for which metadata has changed

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: Yes, we were missing the mock logic for getting it to work. Once I included that, the tests passed

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
jainankitk: Yes, based on the implementation of indicesDeleted, it is not going to be null.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
jainankitk: That is the primary reason for using CounterMetric. It provides eventual consistency without paying synchronization overhead for many concurrent threads. It is used at many places in core for recording the metrics

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
jainankitk: There is no point of adding settings update consumer, as the setting is not dynamic. The cache size cannot be changed dynamically

deshsidd:
Thanks @jainankitk, LGTM overall. Only pending items seem to be:
1. Build fix
2. Removing the TODO and use the settings
3. Unit tests?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
jainankitk: It was failing some tests earlier. Figured out the reason and fixed the tests!

jainankitk: No, I have updated the value to `.1%` of the overall RAM. Let me know if that seems too much or too less

deshsidd (approved):

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -0,0 +1,86 @@
...
+ * Cache implementation specifically for maintaining the field name type mappings
+ * for indices that are part of successful search requests
+ */
+public class IndicesFieldTypeCache {
+
+    private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
+    private final Cache<Index, IndexFieldMap> cache;
+
+    public IndicesFieldTypeCache(Settings settings) {
+        final long sizeInBytes = QueryCategorizationSettings.SEARCH_QUERY_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
```
deshsidd, line 33: No validation needed for the setting?

jainankitk: The validation automatically happens since it is memorySizeSetting

PR #166, opened 2024-12-10
Remove local index custom name setting

dzane17:
### Description
Query Insights allows users to specify a custom index name for storing Top N data via the `search.insights.top_queries.latency.exporter.config.index` cluster setting. However, this flexibility introduces ambiguity. If users change the local index naming pattern, Query Insights will lose track of previously created indices, preventing dashboards from accessing historical data and hindering retention management.

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/exporter/QueryInsightsExporterFactory.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/exporter/QueryInsightsExporterFactory.java
@@ -8,10 +8,8 @@
 
 package org.opensearch.plugin.insights.core.exporter;
 
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_TOP_N_QUERIES_INDEX_PATTERN;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_TOP_QUERIES_EXPORTER_TYPE;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/plugin/insights/core/exporter/QueryInsightsExporterFactory.java
```diff
@@ -73,16 +72,16 @@ public void validateExporterConfig(final Settings settings) throws IllegalArgume
...
         switch (type) {
             case LOCAL_INDEX:
-                final String indexPattern = settings.get(EXPORT_INDEX, DEFAULT_TOP_N_QUERIES_INDEX_PATTERN);
-                if (indexPattern.length() == 0) {
-                    throw new IllegalArgumentException("Empty index pattern configured for the exporter");
-                }
                 try {
-                    DateTimeFormat.forPattern(indexPattern);
+                    DateTimeFormat.forPattern(DEFAULT_TOP_N_QUERIES_INDEX_PATTERN);
                 } catch (Exception e) {
```
ansjcy, line 77: Do we even need to validate this if we are always using defaults?

dzane17: I was wondering about that too. Let's remove it

dzane17: Removed validation from exporter and reader

ansjcy (approved):

Issue #165, opened 2024-12-05
[FEATURE] Better configuration endpoint to control index lifecycle


ansjcy:
### Is your feature request related to a problem?
Query Insights supports rollover historical indices but not deletion. We should enhance Query Insights plugin to better manage query insights index lifecycle, including deletion (and further, adaptive deletion).

### What solution would you like?
* Configuration Endpoint for QI Index Management: In the first phase, we can add a configuration endpoint (and a job) within Query Insights that allows users to define retention policies for their query insights data. By default, the system can be configured to delete query insights data older than 7 days.
* Support for Adaptive Rollover: In the next phase, the configuration endpoint can also enable adaptive rollover based on data volume or configuration, giving us more control over retention without relying on external plugins like ISM.

### What alternatives have you considered?
Relying on ISM for index management, but that means if the ISM plugin faces performance issues or changes, it could impact Query Insights functionality. Customers who disable the ISM plugin might not be able to use this feature neither.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

ansjcy:
We need this feature in 2.19

PR #172, opened 2024-12-18
Top N indices auto deletion config & functionality

dzane17:
### Description
This PR introduces two features related to the retention of Top N historical data:

1. New Cluster Setting: `search.insights.top_queries.delete_after_days`
2. Scheduled job to automatically deletes expired indices

Details:
- `search.insights.top_queries.delete_after_days`: This integer setting determines the retention period in days for Top N indices. Default value is 7 days. Acceptable range is 1 to 180 days.
- Setting is only applicable when `exporter.type` is `local_index`
- Single setting applies to all metric types
- Deletion job runs when `delete_after_days` setting is updated and once per day, in line with the top_queries index rollover period (1 day). Looks for indices that match the naming pattern `top_queries-YYYY.MM.dd-xxxxx`
- Any time `exporter.type` is changed from `local_index`, all existing local indices are deleted. This is so that indices do not linger when sink type is not `local_index`.

### Testing
```
% curl -X GET "localhost:9200/_cat/indices?v&s=index"                                                               
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
yellow open   test2                        I4DeK5ALSOizWRYLbpTGBQ   1   1          4            0      9.2kb          9.2kb
yellow open   top_queries-2025.01.04-62052 kKOZP1ivRlGTtaMP0sXQdw   1   1          7            0     58.1kb         58.1kb
```
```
% curl -XGET "http://localhost:9200/_insights/top_queries?pretty&from=2025-01-01T00:00:00.000Z&to=2025-01-05T00:00:00.000Z"
{
  "top_queries" : [
    {
      "timestamp" : 1735954349215,
      "total_shards" : 1,
      "indices" : [ ],
      "phase_latency_map" : {
        "expand" : 0,
        "query" : 42,
        "fetch" : 2
      },
...
```

`search.insights.top_queries.delete_after_days` validation:
```
% curl -XPUT "http://localhost:9200/_cluster/settings?pretty" -H 'Content-Type: application/json' -d'          
{
  "persistent" : {
    "search.insights.top_queries.delete_after_days" : "999" 
  }
}
'
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "illegal value can't update [search.insights.top_queries.delete_after_days] from [1] to [999]"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "illegal value can't update [search.insights.top_queries.delete_after_days] from [1] to [999]",
    "caused_by" : {
      "type" : "illegal_argument_exception",
      "reason" : "Invalid exporter delete_after_days setting [999], value should be an integer between 1 and 180."
    },
    "suppressed" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "illegal value can't update [search.insights.top_queries.delete_after_days] from [1] to [999]",
        "caused_by" : {
          "type" : "illegal_argument_exception",
          "reason" : "Invalid exporter delete_after_days setting [999], value should be an integer between 1 and 180."
        }
      },
      {
        "type" : "illegal_argument_exception",
        "reason" : "illegal value can't update [search.insights.top_queries.delete_after_days] from [1] to [999]",
        "caused_by" : {
          "type" : "illegal_argument_exception",
          "reason" : "Invalid exporter delete_after_days setting [999], value should be an integer between 1 and 180."
        }
      }
    ]
  },
  "status" : 400
}
```

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/165

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
@@ -8,22 +8,31 @@
 
 package org.opensearch.plugin.insights.core.exporter;
 
+import static org.opensearch.plugin.insights.core.service.TopQueriesService.isTopQueriesIndex;
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_DELETE_AFTER_VALUE;
+
+import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.action.bulk.BulkRequestBuilder;
 import org.opensearch.action.bulk.BulkResponse;
 import org.opensearch.action.index.IndexRequest;
 import org.opensearch.client.Client;
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetric;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetricsCounter;
+import org.opensearch.plugin.insights.core.service.TopQueriesService;
 import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
 
 /**
@@ -36,6 +45,7 @@ public final class LocalIndexExporter implements QueryInsightsExporter {
     private final Logger logger = LogManager.getLogger();
     private final Client client;
     private DateTimeFormatter indexPattern;
+    private int deleteAfter;
 
     /**
      * Constructor of LocalIndexExporter
@@ -46,6 +56,7 @@ public final class LocalIndexExporter implements QueryInsightsExporter {
     public LocalIndexExporter(final Client client, final DateTimeFormatter indexPattern) {
         this.indexPattern = indexPattern;
         this.client = client;
+        this.deleteAfter = DEFAULT_DELETE_AFTER_VALUE;
     }
 
     /**
@@ -61,11 +72,9 @@ public final class LocalIndexExporter implements QueryInsightsExporter {
      * Setter of indexPattern
      *
      * @param indexPattern index pattern
-     * @return the current LocalIndexExporter
      */
-    public LocalIndexExporter setIndexPattern(DateTimeFormatter indexPattern) {
+    void setIndexPattern(DateTimeFormatter indexPattern) {
         this.indexPattern = indexPattern;
-        return this;
     }
 
     /**
@@ -75,15 +84,15 @@ public final class LocalIndexExporter implements QueryInsightsExporter {
      */
     @Override
     public void export(final List<SearchQueryRecord> records) {
-        if (records == null || records.size() == 0) {
+        if (records == null || records.isEmpty()) {
             return;
         }
         try {
-            final String index = getDateTimeFromFormat();
+            final String indexName = buildLocalIndexName();
             final BulkRequestBuilder bulkRequestBuilder = client.prepareBulk().setTimeout(TimeValue.timeValueMinutes(1));
             for (SearchQueryRecord record : records) {
                 bulkRequestBuilder.add(
-                    new IndexRequest(index).source(record.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
+                    new IndexRequest(indexName).source(record.toXContent(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
                 );
             }
             bulkRequestBuilder.execute(new ActionListener<BulkResponse>() {
@@ -110,7 +119,52 @@ public final class LocalIndexExporter implements QueryInsightsExporter {
         logger.debug("Closing the LocalIndexExporter..");
     }
 
-    private String getDateTimeFromFormat() {
-        return indexPattern.format(ZonedDateTime.now(ZoneOffset.UTC));
+    /**
+     * Builds the local index name using the current UTC datetime
+     *
+     * @return A string representing the index name in the format "top_queries-YYYY.MM.dd-01234".
+     */
+    String buildLocalIndexName() {
+        return indexPattern.format(ZonedDateTime.now(ZoneOffset.UTC)) + "-" + generateLocalIndexDateHash();
+    }
+
+    /**
+     * Set local index exporter data retention period
+     *
+     * @param deleteAfter the number of days after which Top N local indices should be deleted
+     */
+    public void setDeleteAfter(final int deleteAfter) {
+        this.deleteAfter = deleteAfter;
+    }
+
+    /**
+     * Delete Top N local indices older than the configured data retention period
+     *
+     * @param indexMetadataMap Map of index name {@link String} to {@link IndexMetadata}
+     */
+    public void deleteExpiredTopNIndices(final Map<String, IndexMetadata> indexMetadataMap) {
+        long expirationMillisLong = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(deleteAfter);
+        for (Map.Entry<String, IndexMetadata> entry : indexMetadataMap.entrySet()) {
+            String indexName = entry.getKey();
+            if (isTopQueriesIndex(indexName) && entry.getValue().getCreationDate() <= expirationMillisLong) {
+                // delete this index
+                TopQueriesService.deleteSingleIndex(indexName, client);
+            }
+        }
+    }
+
+    /**
+     * Generates a consistent 5-digit numeric hash based on the current UTC date.
+     * The generated hash is deterministic, meaning it will return the same result for the same date.
+     *
+     * @return A 5-digit numeric string representation of the current date's hash.
+     */
+    public static String generateLocalIndexDateHash() {
+        // Get the current date in UTC (yyyy-MM-dd format)
+        String currentDate = DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.ROOT)
+            .format(Instant.now().atOffset(ZoneOffset.UTC).toLocalDate());
+
+        // Generate a 5-digit numeric hash from the date's hashCode
+        return String.format(Locale.ROOT, "%05d", (currentDate.hashCode() % 100000 + 100000) % 100000);
     }
 }
--- a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
@@ -10,6 +10,7 @@ package org.opensearch.plugin.insights.core.service;
 
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_GROUPING_TYPE;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR;
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_EXPORTER_DELETE_AFTER;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.getExporterSettings;
 
 import java.io.IOException;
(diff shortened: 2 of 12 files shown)


ansjcy:
Have we done tests on the actual delete functionalities? Since the min value allowed for `delete after` is 1 day it might be hard to test. We can probably temporarily modify the min allowed value to 1 minute and change all the `TimeUnit.Days` to Minutes and test if the deletion work as expected - it should delete all the "top_queries*" indices and shouldn't delete any other indices.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -113,4 +119,47 @@ public void close() {
...
+    public void deleteExpiredIndices(final Map<String, IndexMetadata> indexMetadataMap) {
+        long expirationMillisLong = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(deleteAfter);
+        for (IndexMetadata indexMetadata : indexMetadataMap.values()) {
+            String indexName = indexMetadata.getIndex().getName();
+            if (!matchesPattern(indexName, indexPattern)) {
+                continue;
+            }
+            if (indexMetadata.getCreationDate() <= expirationMillisLong) {
+                // delete this index
+                client.admin().indices().delete(new DeleteIndexRequest(indexMetadata.getIndex().getName()));
```
ansjcy, line 144: `client.admin().indices().delete()` returns a `ActionFuture`, we should capture the delete results, if there are errors/exceptions, we should record the them into an operational metric, or even do some retries. 
something like
```
client.admin().indices().delete(deleteRequest, new ActionListener<>() {
                    @Override
                    public void onResponse(AcknowledgedResponse response) {
                        if (response.isAcknowledged()) {
                            // pass
                        } else {}
                    }

                    @Override
                    public void onFailure(Exception e) {
                       // log in OperationalMetric, retry for one time
                    }
                });
```

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -462,4 +500,13 @@ public QueryInsightsHealthStats getHealthStats() {
             topQueriesHealthStatsMap
         );
     }
+
+    /**
+     * Delete Top N local indices older than the configured data retention period
+     */
+    private void deleteExpiredIndices() {
+        for (MetricType metricType : MetricType.allMetricTypes()) {
+            topQueriesServices.get(metricType).deleteExpiredIndices(clusterService.state().metadata().indices());
```
ansjcy, line 509: for `clusterService.state().metadata().indices()`: I feel this is a little bit dangerous. we are assuming users won't ever have indices named `top_queries-*`. Otherwise we will accidentally delete user data.

I think we should at least add another hash string in the index name, maybe `top_queries-<hash>-YYYY.MM.dd` ? The hash should be computed using a constant so that we can keep the pattern consistent after the OpenSearch process restarts.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -536,4 +540,45 @@ private void drain() {
...
     }
+
+    /**
+     * Validate the exporter delete after value
+     *
+     * @param deleteAfter exporter and reader settings
+     */
+    static void validateExporterDeleteAfter(final int deleteAfter) {
+        if (deleteAfter < MIN_DELETE_AFTER_VALUE || deleteAfter > MAX_DELETE_AFTER_VALUE) {
+            OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.INVALID_EXPORTER_TYPE_FAILURES);
```
ansjcy, line 551: We should use another type instead of reusing `INVALID_EXPORTER_TYPE_FAILURES`

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
ansjcy, line 144: Also on line 144, instead of `indexMetadata.getIndex().getName()`, just use `indexName` ?

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -113,4 +119,47 @@ public void close() {
...
+     * @param input The input string to check.
+     * @param formatter The DateTimeFormatter to validate the string against.
+     * @return true if the string matches the pattern, false otherwise.
+     */
+    static boolean matchesPattern(final String input, final DateTimeFormatter formatter) {
+        try {
+            // Try parsing the input with the given formatter
+            formatter.parseDateTime(input);
+            return true;  // String matches the pattern
+        } catch (IllegalArgumentException e) {
```
ansjcy, line 161: According to https://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormatter.html#parseDateTime-java.lang.String-, `parseDateTime` can also throw [UnsupportedOperationException](https://docs.oracle.com/javase/1.5.0/docs/api/java/lang/UnsupportedOperationException.html?is-external=true). Let's try to catch all the exceptions here.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -113,4 +119,47 @@ public void close() {
...
+     */
+    public void setDeleteAfter(final int deleteAfter) {
+        this.deleteAfter = deleteAfter;
+    }
+
+    /**
+     * Delete Top N local indices older than the configured data retention period
+     */
+    public void deleteExpiredIndices(final Map<String, IndexMetadata> indexMetadataMap) {
+        long expirationMillisLong = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(deleteAfter);
```
ansjcy, line 136: One thought for testing: we can abstract the "finding expired indices" logic from deleteExpiredIndices into a seperate function like
```java
public List<String> findExpiredIndices(final Map<String, IndexMetadata> indexMetadataMap) {
    List<String> expiredIndices = new ArrayList<>();
    long expirationMillisLong = System.currentTimeMillis() - TimeUnit.DAYS.toMillis(deleteAfter);
    
    for (IndexMetadata indexMetadata : indexMetadataMap.values()) {
        String indexName = indexMetadata.getIndex().getName();
        if (matchesPattern(indexName, indexPattern) && 
            indexMetadata.getCreationDate() <= expirationMillisLong) {
            expiredIndices.add(indexName);
        }
    }
    return expiredIndices;
}
```
`deleteExpiredIndices` can then call this function to get the expired indices, we can also add more unit tests on this function by mocking some expired indexMetadata

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -536,4 +540,45 @@ private void drain() {
...
+        if (exporter != null && exporter.getClass() == LocalIndexExporter.class) {
+            ((LocalIndexExporter) exporter).setDeleteAfter(deleteAfter);
+        }
+    }
+
+    /**
+     * Delete Top N local indices older than the configured data retention period
+     */
+    void deleteExpiredIndices(final Map<String, IndexMetadata> indexMetadataMap) {
+        if (exporter != null && exporter.getClass() == LocalIndexExporter.class) {
```
ansjcy, line 579: As I think about this more, should the delete function rely on the index exporter? Cause in this case, if we switch the exporter type, there will be always be lingering indices left since we do the deletion only if `exporter.getClass() == LocalIndexExporter.class`

src/test/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporterTests.java
```diff
@@ -95,4 +97,14 @@ public void testGetAndSetIndexPattern() {
         localIndexExporter.setIndexPattern(newFormatter);
         assert (localIndexExporter.getIndexPattern() == newFormatter);
     }
+
```
ansjcy, line 100: As I mentioned above, we should add more unit tests for deleting the indices here.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
dzane17: On second thought, I am in favor of removing the counter here. This is just validation on a user provided cluster setting value. A spike would only tell us that someone repeatedly provided an invalid input. It does not indicate any defect in the system.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
dzane17: I had the ActionListener but checkstyle would not allow me to import the AcknowledgedResponse class. I saw Query-Insights has some extra github check that is not part of OpenSearch core.

dzane17: Good catch

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
dzane17: I will change it to this:
```
        try {
            // Try parsing the input with the given formatter
            formatter.parseDateTime(input);
            return true;  // String matches the pattern
        } catch (Exception e) {
            return false;  // String does not match the pattern
        }
```
We only care if the string matches so ignoring all exception types.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
dzane17: I have added a UT for `deleteExpiredIndices` with mock index metadata. I will push it once I finish the remaining comments.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
dzane17: That is a good point. Is it okay to perennially check for expired indices when sink type/exporter type is not local index? This involves a for loop over all index names. 

If so, delete logic should move from `LocalIndexExporter` to `TopQueriesService`

src/test/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporterTests.java
dzane17: Added

dzane17:
> Have we done tests on the actual delete functionalities? Since the min value allowed for `delete after` is 1 day it might be hard to test. We can probably temporarily modify the min allowed value to 1 minute and change all the `TimeUnit.Days` to Minutes and test if the deletion work as expected - it should delete all the "top_queries*" indices and shouldn't delete any other indices.

I manually tested with 1 min rollover & 5-10 min retention. It worked as expected.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: I added a 5 digit hash to the index names. The hash is based on the current UTC date so it is consistent across upgrades, node restarts. Let me know if anything else is needed. New index name format: `top_queries-2025.01.04-62052`

ansjcy: Yeah it makes sense to me.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
ansjcy: Since we rollover the index every day, ideally we only need 1 more deletion after disabling / switching off the local index exporter. We need:
- When we turn off local index exporter we can create an one-time job to delete the index next day. 
- When query insights starts, check if there are outdated indices (this is for process restarts.)

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
dzane17: Added `ActionListener` after tip from @ansjcy https://github.com/opensearch-project/security/blob/cd1dcbde45a6c2523505360f2fa15483b170e7f5/src/integrationTest/java/org/opensearch/security/SnapshotSteps.java#L44

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
dzane17: Added a `deleteAllTopNIndices` method in `TopQueriesService`. `deleteAllTopNIndices` is called anytime the `sink_type` setting is changed from `local_index`.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -128,15 +134,22 @@ public QueryInsightsService(
...
+            clusterService.getClusterSettings()
+                .addSettingsUpdateConsumer(
+                    getExporterSettings(type),
+                    (settings -> setExporterAndReader(type, settings, clusterService.state().metadata().indices())),
+                    (settings -> validateExporterAndReaderConfig(type, settings))
+                );
+            clusterService.getClusterSettings()
+                .addSettingsUpdateConsumer(
+                    TOP_N_EXPORTER_DELETE_AFTER,
+                    (settings -> setExporterDeleteAfterAndDelete(type, settings)),
```
ansjcy, line 150: setExporterDeleteAfterAndDelete is only called when delete_after change is modified, should we also invoke it when changing the exporter config?

ansjcy: In https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/core/exporter/QueryInsightsExporterFactory.java#L110

ansjcy (approved):

PR #176, opened 2024-12-23
Migrate from Joda-Time to java.time API

dzane17:
### Description
This PR migrates the codebase from Joda-Time to the `java.time` API to align with Java's standard library and ensure future compatibility, as Joda-Time is no longer maintained. It replaces Joda-Time classes like `DateTime` and `DateTimeZone` with `java.time` equivalents (`ZonedDateTime`, `ZoneOffset`) and updates parsing, formatting, and date-time operations. This change improves maintainability and uses the modern, supported `java.time` API.

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
@@ -8,13 +8,13 @@
 
 package org.opensearch.plugin.insights.core.reader;
 
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-import org.joda.time.format.DateTimeFormatter;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
```diff
@@ -91,18 +91,21 @@ public List<SearchQueryRecord> read(final String from, final String to) {
         if (from == null || to == null) {
             return records;
         }
-        final DateTime start = DateTime.parse(from);
-        DateTime end = DateTime.parse(to);
-        if (end.compareTo(DateTime.now(DateTimeZone.UTC)) > 0) {
-            end = DateTime.now(DateTimeZone.UTC);
+        final ZonedDateTime start = ZonedDateTime.parse(from);
+        ZonedDateTime end = ZonedDateTime.parse(to);
+        ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
```
ansjcy, line 96: I saw local time zones are used in below code like `toLocalDate().atStartOfDay(end.getZone())`, do we also need local timezone here or UTC should also work.

ansjcy (approved):
Looks good, just one minor question. 
Thanks for making the change!

src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
dzane17: All times should be in UTC. `toLocalDate()` is slightly misleading. It does not convert the time. It just extracts the "date" from `end` (ignores "time").

Issue #186, opened 2025-01-14
[BUG] query_group_hashcode is not present in the top_queries response in some cases


deshsidd:
### What is the bug?
_A clear and concise description of the bug._
`query_group_hashcode` is not present in the top_queries response in some cases when the `from` request param is specified.

### How can one reproduce the bug?
_Steps to reproduce the behavior._
1. Configure Top N grouping and local index exporter

2. GET {{BASE_URL}}/_insights/top_queries?type=latency&from=2025-01-13T19%3A56%3A39.870Z&to=2025-01-14T19%3A56%3A39.870Z
```
{
    "top_queries": [
        {
            "timestamp": 1736883631440,
            "labels": {},
            "node_id": "nVSP4I1VRTutynXSQlc5Kw",
            "total_shards": 1,
            "group_by": "SIMILARITY",
            "search_type": "query_then_fetch",
            "phase_latency_map": {
                "expand": 0,
                "query": 29,
                "fetch": 0
            },
            "source": {
                "query": {
                    "match": {
                        "name": {
                            "query": "John",
                            "operator": "OR",
                            "prefix_length": 0,
                            "max_expansions": 50,
                            "fuzzy_transpositions": true,
                            "lenient": false,
                            "zero_terms_query": "NONE",
                            "auto_generate_synonyms_phrase_query": true,
                            "boost": 1.0
                        }
                    }
                }
            },
            "indices": [
                "my-index"
            ],
            "task_resource_usages": [
                {
                    "action": "indices:data/read/search[phase/query]",
                    "taskId": 504,
                    "parentTaskId": 503,
                    "nodeId": "nVSP4I1VRTutynXSQlc5Kw",
                    "taskResourceUsage": {
                        "cpu_time_in_nanos": 17492000,
                        "memory_in_bytes": 2202736
                    }
                },
                {
                    "action": "indices:data/read/search",
                    "taskId": 503,
                    "parentTaskId": -1,
                    "nodeId": "nVSP4I1VRTutynXSQlc5Kw",
                    "taskResourceUsage": {
                        "cpu_time_in_nanos": 837000,
                        "memory_in_bytes": 93336
                    }
                }
            ],
            "measurements": {
                "memory": {
                    "number": 2377288,
                    "count": 3,
                    "aggregationType": "AVERAGE"
                },
                "latency": {
                    "number": 57,
                    "count": 3,
                    "aggregationType": "AVERAGE"
                },
                "cpu": {
                    "number": 23795000,
[...]

Issue #159, opened 2024-11-18
[FEATURE] Add an ID Field to Search Query Records


ansjcy:
### Is your feature request related to a problem?
To support fetching a specific search query record by ID (e.g., for the current detail page or inspecting a historical slow query in the future), we need to add a unique identifier to each search query record.

### What solution would you like?
Add an ID field to search query records.
- Preferred Approach: Use a randomly generated UUID (via UUID.randomUUID() in Java). UUIDs are 128 bits and provide ~3.4 × 10³⁸ possible values, making the probability of collision effectively zero for our use case.

### What alternatives have you considered?
- Use the timestamp as the ID: This may fail when multiple queries share the same timestamp.
- Use a group hash code: While useful for identifying groups, it is insufficient for distinguishing individual queries within a group at different timestamps.
- Generate an ID by hashing the query source and timestamp - this won't avoid ID collision neither

### Do you have any additional context?
This feature is a prerequisite for https://github.com/opensearch-project/query-insights-dashboards/issues/12

ansjcy:
We need to pick this up as part of the QID release.

deshsidd:
As part of this changes need to also incorporate the query_id in the Query Grouping Dashboard PR : https://github.com/opensearch-project/query-insights-dashboards/pull/33

deshsidd:
1. Add query_id
2. Add API for fetching top n for query_id
3. Integrate with QI dashboard

PR #187, opened 2025-01-14
Make sure query_group_hashcode is present in top_queries response in all cases

deshsidd:
### Description
Fixes the bug where `query_group_hashcode` is not present in the top_queries response in some cases when the from request param is specified. This caused duplicate entries on the dashboard and caused all historical rows to be displayed without query_group_hashcode.

Refactoring: Rename `query_hashcode` to `id` as a prereq for https://github.com/opensearch-project/query-insights/issues/159

### Issues Resolved
closes https://github.com/opensearch-project/query-insights/issues/186


ansjcy:
I think we should move the query hash code out from attribute and combine with  https://github.com/opensearch-project/query-insights/issues/159 as an "ID field"

ansjcy (approved):

Issue #146, opened 2024-10-15
[BUG] Add stats for field name to field type mapping cache


jainankitk:
### What is the bug?
As part of https://github.com/opensearch-project/query-insights/pull/142, we added cache for quickly retrieving the field name to field type mapping while ensuring the in-memory size does not grow beyond configured threshold. Currently, we don't have any stats around hits/misses or overall memory in bytes consumed by the cache.

### What is the expected behavior?
We should add these metrics to the health check API of Query Insights plugin

dblock:
[[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta)]

PR #193, opened 2025-01-17
Add field type cache data to health stats API

dzane17:
### Description
Add field type cache stats to the Query Insights `health_stats` API.

This enables users to track total hit count, miss count, entry count, and cache memory size.

```
% curl -XGET "http://localhost:9200/_insights/health_stats?pretty"                     
{
  "WcXrhPnVRdmjADJAFf98aw" : {
    "ThreadPoolInfo" : {
      "query_insights_executor" : {
        "type" : "scaling",
        "core" : 1,
        "max" : 5,
        "keep_alive" : "5m",
        "queue_size" : -1
      }
    },
    "QueryRecordsQueueSize" : 0,
    "TopQueriesHealthStats" : {
      "cpu" : {
        "TopQueriesHeapSize" : 0,
        "QueryGroupCount_Total" : 0,
        "QueryGroupCount_MaxHeap" : 0
      },
      "memory" : {
        "TopQueriesHeapSize" : 0,
        "QueryGroupCount_Total" : 0,
        "QueryGroupCount_MaxHeap" : 0
      },
      "latency" : {
        "TopQueriesHeapSize" : 2,
        "QueryGroupCount_Total" : 2,
        "QueryGroupCount_MaxHeap" : 0
      }
    },
    "FieldTypeCacheStats" : {
      "size_in_bytes" : 336,
      "entry_count" : 3,
      "evictions" : 1,
      "hit_count" : 5,
      "miss_count" : 4
    }
  }
}
```
### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/146

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
@@ -15,10 +15,12 @@ import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.getE
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -36,6 +38,7 @@ import org.opensearch.plugin.insights.core.exporter.QueryInsightsExporterFactory
 import org.opensearch.plugin.insights.core.metrics.OperationalMetric;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetricsCounter;
 import org.opensearch.plugin.insights.core.reader.QueryInsightsReaderFactory;
+import org.opensearch.plugin.insights.core.service.categorizer.QueryShapeGenerator;
 import org.opensearch.plugin.insights.core.service.categorizer.SearchQueryCategorizer;
 import org.opensearch.plugin.insights.rules.model.GroupingType;
 import org.opensearch.plugin.insights.rules.model.MetricType;
@@ -100,9 +103,14 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
 
     private volatile boolean searchQueryMetricsEnabled;
 
-    private SearchQueryCategorizer searchQueryCategorizer;
+    private final SearchQueryCategorizer searchQueryCategorizer;
 
-    private NamedXContentRegistry namedXContentRegistry;
+    private final NamedXContentRegistry namedXContentRegistry;
+
+    /**
+     * Query shape generator instance
+     */
+    private QueryShapeGenerator queryShapeGenerator;
 
     /**
      * Constructor of the QueryInsightsService
@@ -496,10 +504,14 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
         Map<MetricType, TopQueriesHealthStats> topQueriesHealthStatsMap = topQueriesServices.entrySet()
             .stream()
             .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getHealthStats()));
+        Map<String, Long> fieldTypeCacheStats = Optional.ofNullable(queryShapeGenerator)
+            .map(QueryShapeGenerator::getFieldTypeCacheStats)
+            .orElse(Collections.emptyMap());
         return new QueryInsightsHealthStats(
             threadPool.info(QUERY_INSIGHTS_EXECUTOR),
             this.queryRecordsQueue.size(),
-            topQueriesHealthStatsMap
+            topQueriesHealthStatsMap,
+            fieldTypeCacheStats
         );
     }
 
@@ -510,5 +522,12 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
         for (MetricType metricType : MetricType.allMetricTypes()) {
             topQueriesServices.get(metricType).deleteExpiredTopNIndices(clusterService.state().metadata().indices());
         }
+    }
+
+    /**
+     * Set query shape generator
+     */
+    public void setQueryShapeGenerator(final QueryShapeGenerator queryShapeGenerator) {
+        this.queryShapeGenerator = queryShapeGenerator;
     }
 }
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
@@ -28,6 +28,18 @@ public class IndicesFieldTypeCache {
 
     private static final Logger logger = LogManager.getLogger(IndicesFieldTypeCache.class);
     private final Cache<Index, IndexFieldMap> cache;
+    /**
+     * Count of cache evictions
+     */
+    private final CounterMetric evictionCount;
+    /**
+     * Count of items in cache
+     */
+    private final CounterMetric entryCount;
+    /**
+     * Weight of cache in bytes
+     */
+    private final CounterMetric weight;
 
     public IndicesFieldTypeCache(Settings settings) {
         final long sizeInBytes = QueryCategorizationSettings.SEARCH_QUERY_FIELD_TYPE_CACHE_SIZE_KEY.get(settings).getBytes();
@@ -36,9 +48,12 @@ public class IndicesFieldTypeCache {
             cacheBuilder.setMaximumWeight(sizeInBytes).weigher((k, v) -> RamUsageEstimator.sizeOfObject(k) + v.weight());
         }
         cache = cacheBuilder.build();
+        evictionCount = new CounterMetric();
+        entryCount = new CounterMetric();
+        weight = new CounterMetric();
     }
 
-    public IndexFieldMap getOrInitialize(Index index) {
+    IndexFieldMap getOrInitialize(Index index) {
         try {
             return cache.computeIfAbsent(index, k -> new IndexFieldMap());
         } catch (ExecutionException ex) {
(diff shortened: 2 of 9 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -100,9 +105,16 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
...
 
-    private NamedXContentRegistry namedXContentRegistry;
+    private final NamedXContentRegistry namedXContentRegistry;
+
+    /**
+     * Query shape generator instance
+     */
+    private QueryShapeGenerator queryShapeGenerator;
+
+    private static final Map<String, Long> EMPTY_FIELD_TYPE_CACHE_STATS = Map.of(HITS_STRING, 0L, MISSES_STRING, 0L, BYTES_STRING, 0L);
```
jainankitk, line 117: need not be static, should be initialized in the constructor

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -496,10 +508,14 @@ public QueryInsightsHealthStats getHealthStats() {
         Map<MetricType, TopQueriesHealthStats> topQueriesHealthStatsMap = topQueriesServices.entrySet()
             .stream()
             .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getHealthStats()));
+        Map<String, Long> fieldTypeCacheStats = Optional.ofNullable(queryShapeGenerator)
+            .map(QueryShapeGenerator::getFieldTypeCacheStats)
+            .orElse(EMPTY_FIELD_TYPE_CACHE_STATS);
```
jainankitk, line 513: Oh, I see how you using this `EMPTY_FIELD_TYPE_CACHE_STATS`. But in that case, why can't you use Java EMPTY map?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
```diff
@@ -59,9 +59,33 @@ public Iterable<Index> keySet() {
...
+            IndexFieldMap indexFieldMap = cache.get(index);
+
+            // Ensure the value is not null before calling weight()
+            if (indexFieldMap != null) {
+                totalWeight += indexFieldMap.weight();
+            }
+        }
+
+        return totalWeight;
+    }
```
jainankitk, line 80: Shouldn't we compute this while adding or removing from cache?

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -100,9 +105,16 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
...
-    private SearchQueryCategorizer searchQueryCategorizer;
+    private final SearchQueryCategorizer searchQueryCategorizer;
 
-    private NamedXContentRegistry namedXContentRegistry;
+    private final NamedXContentRegistry namedXContentRegistry;
+
+    /**
+     * Query shape generator instance
+     */
+    private QueryShapeGenerator queryShapeGenerator;
```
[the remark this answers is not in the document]
dzane17: Cannot be final. It is defined with a setter method `setQueryShapeGenerator()`, not in the constructor.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: I want to guarantee that `Map<String, Long> fieldTypeCacheStats` contains all 3 keys to avoid serialization issues & missing stats. It should return {hits:0, misses:0, bytes:0} instead of {}.

I think we can pass java empty map here and handle the null case one layer down in QueryInsightsHealthStats though. That will avoid the extra imports. I will change that.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/IndicesFieldTypeCache.java
dzane17: Yes, that makes sense since frequency of `health_stats` API calls will be greater than frequency of items going in/out of this cache.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -38,13 +38,22 @@ public class QueryShapeGenerator implements ClusterStateListener {
...
     private final ClusterService clusterService;
-    private final String NO_FIELD_TYPE_VALUE = "";
     private final IndicesFieldTypeCache indicesFieldTypeCache;
+    private long cacheHitCount;
+    private long cacheMissCount;
+
+    private final String NO_FIELD_TYPE_VALUE = "";
+    public static final String HITS_STRING = "hits";
+    public static final String MISSES_STRING = "misses";
+    public static final String BYTES_STRING = "bytes";
```
jainankitk, line 48: Should we have stat for number of entries as well?

src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryInsightsHealthStats.java
```diff
@@ -53,14 +62,16 @@ public QueryInsightsHealthStats(final StreamInput in) throws IOException {
...
+        final Map<MetricType, TopQueriesHealthStats> topQueriesHealthStats,
+        final Map<String, Long> fieldTypeCacheStats
     ) {
         if (threadPoolInfo == null || topQueriesHealthStats == null) {
             throw new IllegalArgumentException("Parameters cannot be null");
         }
         this.threadPoolInfo = threadPoolInfo;
         this.queryRecordsQueueSize = queryRecordsQueueSize;
         this.topQueriesHealthStats = topQueriesHealthStats;
+        this.fieldTypeCacheStats = fieldTypeCacheStats != null ? fieldTypeCacheStats : ZEROS_FIELD_TYPE_CACHE_STATS;
```
jainankitk, line 74: Why can't this be empty map?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -369,14 +378,20 @@ String getFieldType(String fieldName, Map<String, Object> propertiesAsMap, Index
...
         }
 
         // Retrieve field type from mapping and cache it if found
         fieldType = getFieldTypeFromProperties(fieldName, propertiesAsMap);
 
         // Cache field type or NO_FIELD_TYPE_VALUE if not found
-        indicesFieldTypeCache.getOrInitialize(index).putIfAbsent(fieldName, fieldType != null ? fieldType : NO_FIELD_TYPE_VALUE);
+        fieldType = fieldType != null ? fieldType : NO_FIELD_TYPE_VALUE;
+        if (indicesFieldTypeCache.getOrInitialize(index).putIfAbsent(fieldName, fieldType)) {
+            indicesFieldTypeCache.incrementWeight(fieldName, fieldType);
```
jainankitk, line 393: Isn't the weight incremented twice, during `putIfAbsent` and also during `incrementWeight`?

dzane17: There is a weight for the overall cache `IndicesFieldTypeCache` and individual weights for the per-index maps `IndexFieldMap`

src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryInsightsHealthStats.java
dzane17: I am concerned about streaming data between nodes. We are using read map in & out so should be able to handle missing keys.

Default value in `toXContent()` ensures output always contains desired keys. 
```
        // Write field type cache stats
        builder.startObject(FIELD_TYPE_CACHE_STATS);
        for (String key : List.of("hits", "misses", "bytes")) {
            builder.field(key, fieldTypeCacheStats.getOrDefault(key, 0L));
        }
```

I will do some testing to verify

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
dzane17: Added count

src/main/java/org/opensearch/plugin/insights/rules/model/healthStats/QueryInsightsHealthStats.java
dzane17: Changed to empty map

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/QueryShapeGenerator.java
```diff
@@ -38,13 +38,22 @@ public class QueryShapeGenerator implements ClusterStateListener {
...
-    private final String NO_FIELD_TYPE_VALUE = "";
     private final IndicesFieldTypeCache indicesFieldTypeCache;
+    private long cacheHitCount;
+    private long cacheMissCount;
+
+    private final String NO_FIELD_TYPE_VALUE = "";
+    public static final String HITS_STRING = "hits";
+    public static final String MISSES_STRING = "misses";
+    public static final String COUNT_STRING = "count";
+    public static final String BYTES_STRING = "bytes";
```
jainankitk, line 49: Can we rename these to below for better clarity and consistency? And, also add eviction metric as well!?

```
"size_in_bytes": 4096,
"entry_count": 1000,
"evictions": 2,
"hit_count": 60,
"miss_count": 10
```

dzane17: Done

jainankitk (approved):

Issue #204, opened 2025-01-24
[BUG] Error thrown when parsing local index data


ansjcy:
### What is the bug?
Currently we are not collecting metrics that are not enabled for top n queries, which causes certain measurements fields missing in local index sometimes. This is not a consistent behavior, idealy, when user configure "enabling/disabling top n queries by cpu", we should really just adding/removing the logic to push the records into the cpu metric heap, but whether the the top n queries are configured or not, we should always `collect` the metric.

### How can one reproduce the bug?
- Enable top n queries by latency only
- Enable local index exporter
- Wait untail the exporter exports to local index
- Enable top n queries by memory or cpu
- try to get historical top n queries with `from` and `to`

### What is the expected behavior?
No error should be thrown and we should always collect the metrics

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

PR #205, opened 2025-01-24
Always collect available metrics in top queries service

ansjcy:
### Description
Currently we are not collecting metrics that are not enabled for top n queries, which causes certain measurements fields missing in local index sometimes. This is not a consistent behavior, idealy, when user configure "enabling/disabling top n queries by cpu", we should really just adding/removing the logic to push the records into the cpu metric heap, but whether the the top n queries are configured or not, we should always collect the metric.

### Issues Resolved
https://github.com/opensearch-project/query-insights/issues/204


deshsidd (approved):
LGTM. Good find!

PR #210, opened 2025-01-24
Refactor Exporters and Readers

ansjcy:
### Description
There are multiple issues with the current exporter.

1. Currently, we have different exporters config for different metrics.
https://opensearch.org/docs/latest/observing-your-data/query-insights/top-n-queries/#configuring-a-local-index-exporter
This is not a good pattern since we don't have specific use cases when users enabled multiple top n metrics, but they only want to exporter certain of them. We need to move the exporter from `top queries service` level to `query insights service` level.
2. The safeguard to decide if an index is created by query insights service is not strong. We only check the index name. We should add certain metadata in the index to indicate "this is our index". https://github.com/opensearch-project/query-insights/issues/213
3. There's no index mapping and index level sorting defined for the index, and we only allow default size (10?) when searching on the index. This will become a problem when the frontend wants lots of data (basically, everything exceed 10 will go wrong) https://github.com/opensearch-project/query-insights/issues/212
4. There are a lot of duplicate records in index when multiple metrics are enabled. We need to do deduplication.
5. when LocalIndexReader uses generateLocalIndexDateHash() to get historical data, since generateLocalIndexDateHash only consider current time, it's not generating correct hash for older date.
6. We are overusing static methods, and not complying with the ExporterFactory Model. Need to refactor the node.

### Issues Resolved
TODO

### tests
TODO
- enable top n queries for all metrics and local index exporter with
```
 curl -X PUT 'localhost:9200/_cluster/settings' -H 'Content-Type: application/json' -d'
{
    "persistent" : {
        "search.insights.top_queries.latency.enabled" : "true",
        "search.insights.top_queries.cpu.enabled" : "true",
        "search.insights.top_queries.memory.enabled" : "true",
        "search.insights.top_queries.exporter.type" : "local_index"
    }
}'
{"acknowledged":true,"persistent":{"search":{"insights":{"top_queries":{"cpu":{"enabled":"true"},"exporter":{"type":"local_index"},"memory":{"enabled":"true"},"latency":{"enabled":"true"}}}}},"transient":{}}
```
- do search
- check local index, data are exported without errors
```
➜  OpenSearch git:(main) ✗ curl -X POST "localhost:9200/top_queries-*/_search?pretty"
{
  "took" : 15,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 6,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "top_queries-2025.01.24-62114",
        "_id" : "l1L7lZQBbSnBp-lpBZtY",
        "_score" : 1.0,
        "_source" : {
          "timestamp" : 1737683015018,
          "id" : "ba5087e8-47f6-4882-b4e0-94b0c455df83",
          "task_resource_usages" : [
            {
              "action" : "indices:data/read/search[phase/query]",
              "taskId" : 117,
              "parentTaskId" : 116,
              "nodeId" : "aBdathzfR_O-o7ct8LqHrA",
              "taskResourceUsage" : {
                "cpu_time_in_nanos" : 577000,
                "memory_in_bytes" : 33160
              }
            },
            {
              "taskId" : 116,
              "parentTaskId" : -1,
              "nodeId" : "aBdathzfR_O-o7ct8LqHrA",
              "taskResourceUsage" : {
                "cpu_time_in_nanos" : 85000,
                "memory_in_bytes" : 3672
              }
            }
          ],
          "search_type" : "query_then_fetch",
          "total_shards" : 1,
          "indices" : [
            "my-index-0"
          ],
          "group_by" : "NONE",
          "labels" : { },
          "phase_latency_map" : {
            "expand" : 0,
            "query" : 1,
            "fetch" : 0
          },
          "source" : {
            "size" : 20,
            "query" : {
              "term" : {
                "user.id" : {
                  "value" : "cyji",
                  "boost" : 1.0
                }
              }
            }
          },
          "measurements" : {
            "memory" : {
              "number" : 36832,
              "count" : 1,
              "aggregationType" : "NONE"
            },
            "latency" : {
              "number" : 1,
              "count" : 1,
              "aggregationType" : "NONE"
            },
            "cpu" : {
              "number" : 662000,
              "count" : 1,
              "aggregationType" : "NONE"
            }
          }
        }
      },
    ...
```


src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -138,33 +220,58 @@ public void setDeleteAfter(final int deleteAfter) {
...
+    /**
+     * check if index exists
+     * @return boolean
      */
-    public static String generateLocalIndexDateHash() {
-        // Get the current date in UTC (yyyy-MM-dd format)
-        String currentDate = DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.ROOT)
-            .format(Instant.now().atOffset(ZoneOffset.UTC).toLocalDate());
+    private boolean checkIndexExists(String indexName) {
+        ClusterState clusterState = clusterService.state();
```
deshsidd, line 261: Can there be a corner scenario where the cluster state does not have up to date information? Are we handling this?

src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
```diff
@@ -104,7 +116,7 @@ public List<SearchQueryRecord> read(final String from, final String to, String i
         while (curr.isBefore(end.plusDays(1).toLocalDate().atStartOfDay(end.getZone()))) {
             String indexName = buildLocalIndexName(curr);
             SearchRequest searchRequest = new SearchRequest(indexName);
-            SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+            SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder().size(1000);
```
deshsidd, line 119: Might want to add a constant so this does not get lost in the code in case we want to change this in the future

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -407,28 +416,68 @@ public void setTopNSize(final MetricType type, final int topNSize) {
...
-            TopQueriesService tqs = topQueriesServices.get(type);
-            tqs.setExporter(settings, indexMetadataMap);
-            tqs.setReader(settings, namedXContentRegistry);
+    private void setExporterAndReader(final SinkType sinkType, final Map<String, IndexMetadata> indexMetadataMap) {
+        final QueryInsightsExporter topQueriesExporter = queryInsightsExporterFactory.getExporter(TOP_QUERIES_LOCAL_INDEX_EXPORTER_ID);
+
+        // This method is invoked when sink type is changed
+        // Clear local indices if exporter is of type LocalIndexExporter
+        if (topQueriesExporter != null && topQueriesExporter.getClass() == LocalIndexExporter.class) {
+            deleteAllTopNIndices(client, indexMetadataMap, (LocalIndexExporter) topQueriesExporter);
```
deshsidd, line 428: If we toggle between exports and back to local index exporter do we want to lose all the older data in this scenario?

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -592,89 +519,55 @@ static void validateExporterDeleteAfter(final int deleteAfter) {
...
+     * @param indexMetadata the metadata associated with the index
+     * @return {@code true} if the string is valid, {@code false} otherwise.
      */
-    void deleteExpiredTopNIndices(final Map<String, IndexMetadata> indexMetadataMap) {
-        if (exporter != null && exporter.getClass() == LocalIndexExporter.class) {
-            threadPool.executor(QUERY_INSIGHTS_EXECUTOR)
-                .execute(() -> ((LocalIndexExporter) exporter).deleteExpiredTopNIndices(indexMetadataMap));
-        }
-    }
+    public static boolean isTopQueriesIndex(String indexName, IndexMetadata indexMetadata) {
```
deshsidd, line 529: nit: Do we want to add this to a helper/util?

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -222,23 +221,15 @@ public class QueryInsightsSettings {
...
-    private static final String TOP_N_LATENCY_QUERIES_EXPORTER_PREFIX = TOP_N_LATENCY_QUERIES_PREFIX + ".exporter.";
-    /**
-     * Prefix for top n queries by cpu exporters
-     */
-    private static final String TOP_N_CPU_QUERIES_EXPORTER_PREFIX = TOP_N_CPU_QUERIES_PREFIX + ".exporter.";
-    /**
-     * Prefix for top n queries by memory exporters
-     */
-    private static final String TOP_N_MEMORY_QUERIES_EXPORTER_PREFIX = TOP_N_MEMORY_QUERIES_PREFIX + ".exporter.";
+    private static final String TOP_N_QUERIES_EXPORTER_PREFIX = TOP_N_QUERIES_SETTING_PREFIX + ".exporter";
```
deshsidd, line 224: Self note: Will have to update the documentation since the current docs are now invalid.

src/test/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporterTests.java
```diff
@@ -116,32 +186,32 @@ public void testGetAndSetIndexPattern() {
...
-            indexMetadataMap.put(indexName, indexMetadata);
-        }
-        localIndexExporter.deleteExpiredTopNIndices(indexMetadataMap);
-        // Default retention is 7 days
-        // Oldest 3 of 10 indices should be deleted
-        verify(client, times(3)).admin();
-        verify(adminClient, times(3)).indices();
-        verify(indicesAdminClient, times(3)).delete(any(), any());
-    }
+    // public void testDeleteExpiredTopNIndices() {
```
deshsidd, line 189: Why is this commented out?

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
```diff
@@ -224,4 +252,115 @@ public void testGetHealthStats() {
...
+                .build();
+            indexMetadataMap.put(indexName, indexMetadata);
+        }
+
+        queryInsightsService.deleteAllTopNIndices(client, indexMetadataMap, localIndexExporter);
+        // All 10 indices should be deleted
+        verify(localIndexExporter, times(9)).deleteSingleIndex(any(), any());
+    }
+
+    // TODO: comment out for now. Need to reenable this before mering
```
deshsidd, line 302: nit: Don't forget to resolve this

deshsidd (commented):
Added a few comments and went over this - looks good at a high level. Might need to give one more pass soon.

src/test/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporterTests.java
ansjcy: It is moved to queryInsightsTests.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
ansjcy: Yes, In theory the cluster state should be most up to date but there are scenarios when 2 threads are trying to create / delete the same index for the same time. If you look into the code I already handled with something like
```
Throwable cause = ExceptionsHelper.unwrapCause(e);
                        if (cause instanceof ResourceAlreadyExistsException) {
```

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
ansjcy: Yes, that's the expectation for now. Disabling local index exporter will delete all historical data. It was originally implemented in https://github.com/opensearch-project/query-insights/pull/172

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
ansjcy: I think it should belongs to TopQueriesService though since we are checking if an index belongs to top n queries.

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
ansjcy: yes, I'll send a PR to update the document later.

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
ansjcy: Updated

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -138,33 +221,58 @@ public void setDeleteAfter(final int deleteAfter) {
...
+        return clusterState.getRoutingTable().hasIndex(indexName);
```
deshsidd, line 263: Hope we don't need to handle any potential errors here

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -407,28 +416,68 @@ public void setTopNSize(final MetricType type, final int topNSize) {
...
+                queryInsightsExporterFactory.createExporter(
+                    TOP_QUERIES_LOCAL_INDEX_EXPORTER_ID,
+                    sinkType,
+                    DEFAULT_TOP_N_QUERIES_INDEX_PATTERN,
+                    "mappings/top-queries-record.json"
+                );
+            }
+        } else {
+            // Disable exporter if exporter type is set to null
+            try {
```
deshsidd, line 453: nit: Duplicated code few lines above, could refactor this out

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
```diff
@@ -224,4 +281,128 @@ public void testGetHealthStats() {
...
+                )
+                .putMapping(
+                    new MappingMetadata("_doc", Map.of("_meta", Map.of(QUERY_INSIGHTS_INDEX_TAG_NAME, TOP_QUERIES_INDEX_TAG_VALUE)))
+                )
+                .build();
+            indexMetadataMap.put(indexName, indexMetadata);
+        }
+
+        queryInsightsService.deleteAllTopNIndices(client, indexMetadataMap, localIndexExporter);
+        // All 10 indices should be deleted
```
deshsidd, line 330: 9 indices should be deleted?

deshsidd (approved):
LGTM overall with some minor comments.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
ansjcy: Good point.. I saw the same patter is used in multiple places in OpenSearch. I think we should be good.
https://github.com/opensearch-project/OpenSearch/blob/e6fc6008e408c41e2369ca308d5a763575ef8c89/server/src/main/java/org/opensearch/cluster/health/ClusterStateHealth.java#L211

ansjcy:
@deshsidd Thanks for the review, I'll merge it for now to unblock the frontend, but we need to iterate from here to add missing doc/ refactor some code into util functions etc.

PR #223, opened 2025-01-29
Fix grouping integ tests

dzane17:
### Description
Resolves https://github.com/opensearch-project/query-insights/issues/220


deshsidd:
Thanks! This might work since the window size will be set to 1m and then we are waiting for about 70 seconds. Hope no local index historical data is coming into play here. Don't think we have ITs for this.

deshsidd (approved):

Issue #261, opened 2025-03-10
[FEATURE] Better safeguard when deleting outdated query insights indices


ansjcy:
### Is your feature request related to a problem?
The current implementation uses the [`isTopQueriesIndex` function](https://github.com/opensearch-project/query-insights/blame/5ef54373e9eba9c9e0fbe80090e77c7e0953960a/src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java#L516)  to filter indices when deleting outdated data. While this approach works currently, it presents a potential risk as the codebase evolves. If we modify index settings or patterns and update this function without careful consideration, we might inadvertently delete data from unrelated indices, as we currently iterate through all indices in the cluster.

### What solution would you like?
Replace the broad `clusterService.state().metadata().indices()` approach to get delete candidates with a more targeted method that only considers indices matching our specific pattern, something like:
```java
 ClusterStateRequest clusterStateRequest = new ClusterStateRequest()
            .clear()
            .indices(indexPattern)
            .metadata(true)
            .local(true)
            .indicesOptions(IndicesOptions.strictExpand());
```
This can limit deletion candidates exclusively to indices matching the top queries pattern, creating a safer boundary for our deletion operations.

### What alternatives have you considered?
We could maintain the current implementation, but this would perpetuate the risk of potential data loss as the software evolves, particularly during refactoring or feature expansion.

### Do you have any additional context?
This change follows the principle of least privilege by limiting the scope of our deletion operations to only the indices we intend to manage.

PR #262, opened 2025-03-10
Use ClusterStateRequest with index pattern when searching for expired local indices

dzane17:
### Description
Currently we scan a list of all indices on a domain when looking for expired Top N local indices. This causes extra computation on domains with many indices & greater possibility of mistake.

This PR reduces risk by limiting the indices fetched to those that match the `top_queries-*` name pattern. It then applies the existing protocol, which includes a stricter check for the index name (`top_queries-<date>-<5 digits>`), as well as validation of creation time and metadata before any deletions are performed.

### Testing
1. Create a few indices including 1 real Query Insights local index
```
% curl -X GET "localhost:9200/_cat/indices?v&s=index"                                                       
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   logs-1                       LUay8iGnTZSSW3rjMUHiKA   1   0          0            0       208b           208b
green  open   logs-2                       I1y6WAWkSF2mTy8QtX1Pcg   1   0          0            0       208b           208b
yellow open   my_index                     h27P91J7TNCdv1rMhFB6fg   1   1          2            0      6.4kb          6.4kb
green  open   top_queries-2020.01.01-12345 _AaDcAN4TEujGrFMjGLidw   1   0          0            0       208b           208b
green  open   top_queries-2025.03.10-21661 cyi2BDaLS42BlO6lmwpO9Q   1   0          1            2       29kb           29kb
```

2. Update `delete_after_days` setting to trigger expired index scan + deletion
```
% curl -XPUT "http://localhost:9200/_cluster/settings?pretty" -H 'Content-Type: application/json' -d'          
{              
  "persistent" : {        
    "search.insights.top_queries.exporter.delete_after_days" : "5"
  }
} 
'
{
  "acknowledged" : true,
  "persistent" : {
    "search" : {
      "insights" : {
        "top_queries" : {
          "exporter" : {
            "delete_after_days" : "5"
          }
        }
      }
    }
  },
  "transient" : { }
}
```

3. Verified with logging, only indices with name matching `top_queries-*` pattern were considered for deletion
```
[2025-03-10T16:24:45,187][INFO ][o.o.c.s.ClusterSettings  ] [integTest-0] updating [search.insights.top_queries.exporter.delete_after_days] from [7] to [5]
[2025-03-10T16:24:45,188][ERROR][o.o.p.i.c.s.QueryInsightsService] [integTest-0] in deleteExpiredTopNIndices
[2025-03-10T16:24:45,188][ERROR][o.o.p.i.c.s.QueryInsightsService] [integTest-0] top_queries-2020.01.01-12345
[2025-03-10T16:24:45,189][ERROR][o.o.p.i.c.s.QueryInsightsService] [integTest-0] top_queries-2025.03.10-21661
```

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/261

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
@@ -18,6 +18,7 @@ import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.MIN_
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_EXPORTER_DELETE_AFTER;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_EXPORTER_TYPE;
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_QUERIES_INDEX_PATTERN_GLOB;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -33,11 +34,14 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
+import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.lifecycle.AbstractLifecycleComponent;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.plugin.insights.core.exporter.LocalIndexExporter;
 import org.opensearch.plugin.insights.core.exporter.QueryInsightsExporter;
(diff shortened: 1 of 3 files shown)


ansjcy:
I think you also need to make the same changes here: https://github.com/opensearch-project/query-insights/blob/5ef54373e9eba9c9e0fbe80090e77c7e0953960a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java#L622

ansjcy:
Another improvements we can do is, on this line https://github.com/opensearch-project/query-insights/blob/5ef54373e9eba9c9e0fbe80090e77c7e0953960a/src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java#L555 instead of just `parts[2].matches("\\d{5}");`, we can call this function to verify if the hash is correct as well.
https://github.com/opensearch-project/query-insights/blob/5ef54373e9eba9c9e0fbe80090e77c7e0953960a/src/main/java/org/opensearch/plugin/insights/core/utils/ExporterReaderUtils.java#L29

Let me know if you want to include it in this PR as well, if not I'll probably do it in another PR.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -598,17 +602,25 @@ void deleteExpiredTopNIndices() {
...
-                );
-                for (Map.Entry<String, IndexMetadata> entry : indexMetadataMap.entrySet()) {
-                    String indexName = entry.getKey();
-                    if (isTopQueriesIndex(indexName, entry.getValue()) && entry.getValue().getCreationDate() <= expirationMillisLong) {
-                        // delete this index
-                        localIndexExporter.deleteSingleIndex(indexName, client);
+                final ClusterStateRequest clusterStateRequest = new ClusterStateRequest().clear()
+                    .indices(TOP_QUERIES_INDEX_PATTERN_GLOB)
+                    .metadata(true)
+                    .local(true)
```
ansjcy, line 608: According to the java doc, the local parameter is defined as: "Return local information, do not retrieve the state from cluster-manager node (default: false).". 

using data from cluster manager node requires one extra network hop, it will also have higher latency, but the benifit is we will always have the most accurate info..

But in our case, the absolute consistency is not critical neither.. I'm fine with either approach, just leave the info here in case anyone has other opinions.

dzane17:
I will update `deleteAllTopNIndices()` and the 5 digit check in `isTopQueriesIndex()`.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Is cluster state equivalent on cluster manager vs non cluster manager nodes? I am wondering if local node info can be incomplete or stale.

ansjcy (approved):

Issue #288, opened 2025-03-26
[BUG] Improve delete after execution timing


dzane17:
### What is the bug?
Local indices are retained longer than expected. For example, we observed a domain containing local indices for the past 3 days when `delete_after_days=1`.

This is caused by 2 factors:
1. Periodic job to delete indices is scheduled at a 24 hours interval starting from when the domain/node is created. If a node boots up at 14:30, the job will run at 14:30 every day in the future. We have no control over when the OpenSearch process starts, so this is non deterministic.
https://github.com/opensearch-project/query-insights/blob/a87649568773fb3f03daaa221fc740171892fdef/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java#L581-L587
2. Expiration is calculated with current time. That means the local index creation time in relation to the daily job start time can add an extra day of retention.
https://github.com/opensearch-project/query-insights/blob/a87649568773fb3f03daaa221fc740171892fdef/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java#L649-L651

### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
Do not allow local indices to overstay the delete_after_days setting value

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

PR #289, opened 2025-03-27
Fix local index deletion timing

dzane17:
### Description
Fix local index deletion
- Run job every day at 00:05 UTC instead of when `QueryInsightsService` was launched
- Use 00:00 as expiration cutoff instead of `current_time - delete_after_days`
- Check & delete expired indices on cluster manager node only instead of all nodes

### Testing
Added logs to validate:
1. `getInitialDelay()` returns correct duration
2. `deleteExpiredTopNIndices()` is called at 00:05 UTC

```
[2025-04-08T23:59:24,998][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] scheduling job
[2025-04-08T23:59:24,999][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] in getInitialDelay
[2025-04-08T23:59:24,999][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] delay = 35000
[2025-04-08T23:59:24,999][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] scheduled job
...
[2025-04-09T00:05:00,000][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] starting deleteExpiredTopNIndices
```

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/288

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
@@ -22,6 +22,11 @@ import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_QUERIES_INDEX_PATTERN_GLOB;
 
 import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -31,6 +36,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
@@ -38,10 +44,12 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.action.admin.cluster.state.ClusterStateRequest;
 import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.node.DiscoveryNodeRole;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.lifecycle.AbstractLifecycleComponent;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.common.util.concurrent.FutureUtils;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.plugin.insights.core.exporter.LocalIndexExporter;
@@ -101,6 +109,11 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
      * the service closed concurrently.
      */
     protected volatile List<Scheduler.Cancellable> scheduledFutures;
+
+    /**
+     * Reference to the Query Insights local index deletion {@link ScheduledFuture}
+     */
+    protected volatile ScheduledFuture deleteIndicesScheduledFuture;
 
     /**
      * Factory for validating and creating exporters
@@ -567,6 +580,21 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
         }
     }
 
+    /**
+     * Calculates the duration until the next UTC midnight (00:00:00).
+     * <p>
+     * This method computes the time difference between the given {@code Instant} and the start of the next UTC day
+     * at midnight (00:00:00). It returns the duration as a long in milliseconds.
+     *
+     * @param now The current time as an {@link Instant}.
+     * @return A long representing the time delay (in milliseconds) until the next UTC midnight.
+     */
+    static long getInitialDelay(final Instant now) {
+        // Calculate the start of the next UTC day (00:00:00)
+        final Instant startOfNextDay = now.truncatedTo(ChronoUnit.DAYS).plus(1, ChronoUnit.DAYS);
+        return Duration.between(now, startOfNextDay).toMillis();
+    }
+
     @Override
     protected void doStart() {
         if (isAnyFeatureEnabled()) {
@@ -578,13 +606,20 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
                     QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR
                 )
             );
-            scheduledFutures.add(
-                threadPool.scheduleWithFixedDelay(
-                    this::deleteExpiredTopNIndices,
-                    new TimeValue(1, TimeUnit.DAYS), // Check for deletable indices once per day
-                    QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR
-                )
-            );
+            if (threadPool.scheduler() != null) {
+                deleteIndicesScheduledFuture = threadPool.scheduler().scheduleWithFixedDelay(() -> {
+                    try {
+                        if (clusterService.isStateInitialised()
+                            && clusterService.localNode().getRoles().contains(DiscoveryNodeRole.CLUSTER_MANAGER_ROLE)) {
+                            deleteExpiredTopNIndices();
+                        }
+                    } catch (Exception e) {
+                        logger.error("Error occurred while deleting expired indices:", e);
+                    }
+                }, getInitialDelay(Instant.now()) + Duration.ofMinutes(5).toMillis(), Duration.ofDays(1).toMillis(), TimeUnit.MILLISECONDS);
+            } else {
+                logger.error("Unable to schedule Query Insights delete job. threadPool.scheduler() is null");
+            }
         }
     }
 
@@ -597,6 +632,7 @@ public class QueryInsightsService extends AbstractLifecycleComponent {
                 }
             }
         }
+        FutureUtils.cancel(deleteIndicesScheduledFuture);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


deshsidd:
LGTM

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
```diff
@@ -559,6 +562,20 @@ public void testSetExporterAndReaderType_CloseWithException() throws IOException
...
+        Instant instantNormal = Instant.parse("2025-03-26T10:15:30Z");
+        Instant startOfNextDayNormal = instantNormal.truncatedTo(ChronoUnit.DAYS).plus(1, ChronoUnit.DAYS);
+        final long expectedDelayNormal = Duration.between(instantNormal, startOfNextDayNormal).toMillis();
+        assertEquals(expectedDelayNormal, getInitialDelay(instantNormal).millis());
+
+        //// Second Test Case (Edge Case: Midnight UTC) ////
+        Instant instantEdge = Instant.parse("2025-03-26T00:00:00Z");
+        Instant startOfNextDayEdge = instantEdge.plus(1, ChronoUnit.DAYS);
+        final long expectedDelayEdge = Duration.between(instantEdge, startOfNextDayEdge).toMillis();
+        assertEquals(expectedDelayEdge, getInitialDelay(instantEdge).millis());
```
ansjcy, line 576: Is it possible to add tests for schedule after the first day by mocking the scheduler?

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -567,6 +572,41 @@ public void validateTemplatePriority(final long templatePriority) {
...
     }
 
+    /**
+     * Schedules a daily task to delete expired Top N indices.
+     *
+     * @see #deleteExpiredTopNIndices()
+     */
+    private void createDailyDeleteAfterJob() {
+        // Run job once now
+        deleteExpiredTopNIndices();
```
ansjcy, line 582: Could you add some error handling here to avoid crashing the process if error happens? If `deleteExpiredTopNIndices` somehow error out, we should also be able to add the schedule job.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -567,6 +572,41 @@ public void validateTemplatePriority(final long templatePriority) {
...
+
+        // Schedule recurring job with 1-day interval
+        scheduledFutures.add(
```
ansjcy, line 585: Just had a thought, this QueryInsightsService runs on every node, we're basically spinning up N deletion jobs right now. This is an issue in the currently implementation as well. We should really just create 1 job here.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -567,6 +572,41 @@ public void validateTemplatePriority(final long templatePriority) {
...
+     */
+    static TimeValue getInitialDelay(final Instant now) {
+        // Calculate the start of the next UTC day (00:00:00)
+        final Instant startOfNextDay = now.truncatedTo(ChronoUnit.DAYS).plus(1, ChronoUnit.DAYS);
+        final long delayInMillis = Duration.between(now, startOfNextDay).toMillis();
+        return TimeValue.timeValueMillis(delayInMillis);
+    }
+
     @Override
     protected void doStart() {
```
ansjcy, line 611: Looks like we don't have any unit tests for the doStart method yet. Since we're adding this createDailyDeleteAfterJob stuff which makes things more complex, we should probably add some tests to cover it. That way we can make sure threadPool.schedule gets called properly, and if someone changes something later that breaks it, the tests will catch it.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -578,12 +618,12 @@ protected void doStart() {
...
-                threadPool.scheduleWithFixedDelay(
-                    this::deleteExpiredTopNIndices,
-                    new TimeValue(1, TimeUnit.DAYS), // Check for deletable indices once per day
-                    QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR
-                )
+
+            // One-time job to run at next UTC 00:00
+            threadPool.schedule(
+                this::createDailyDeleteAfterJob,
+                getInitialDelay(Instant.now()),
```
ansjcy, line 625: The potential problem of this approach is, think about 2 cases (say today is April 2nd and delete after is set to 1 day):
- if the process is started at 23:59:59:98, it's possible `createDailyDeleteAfterJob` will be scheduled on 00:00 on the next day, then deleteExpiredTopNIndices will delete everything before April 1
- if the process is started at 23:59:59:98, it's also possible `createDailyDeleteAfterJob` will be scheduled on 23:59:59:99 on the same day, then deleteExpiredTopNIndices will delete everything before March 31.

Then the behavior might be inconsistent. Instead of `// One-time job to run at next UTC 00:00`, can we schedule it at some other time to make the behavior consistent?

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -567,6 +572,41 @@ public void validateTemplatePriority(final long templatePriority) {
...
+     *
+     * @see #deleteExpiredTopNIndices()
+     */
+    private void createDailyDeleteAfterJob() {
+        // Run job once now
+        deleteExpiredTopNIndices();
+
+        // Schedule recurring job with 1-day interval
+        scheduledFutures.add(
+            threadPool.scheduleWithFixedDelay(
```
ansjcy, line 586: We need tests to make sure this path is called by mocking the system time.

ansjcy (changes requested):

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Added try/catch around `deleteExpiredTopNIndices()`

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Added a test for doStart

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Changed execution time from 00:00 to 00:05. That way we are clear of any rounding/millisecond errors that could cause it to run on the previous day.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Good point. We should only schedule the delete job on the cluster manager node. With that said, we need to ensure the job stops/starts anytime the following events occur. 1-2 are already handled in the `QueryInsightsService` `doStart()` method. For number 3 I have added a listener which is called anytime the cluster manager node changes.
1. QI process restart
2. All top N & search metrics features toggle off/on
3. Cluster manager node changes

Another piece of complexity is that `QueryInsightsService` is created before `clusterService.clusterApplierService.state()` is initialized. This means the cluster manager node is unknown when QI process starts. Deletion job in this case will be scheduled by the listener once state is defined. If this order were to change, we are still okay. The job will be scheduled in `QueryInsightsService.doStart()` and the listener will pass since there is already an active ScheduledFuture present.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: New `testDoStart()` checks a ScheduledFuture is created.

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
dzane17: I don't see a way to truly mock the passage of time and ensure `deleteExpiredTopNIndices` is called at a certain time by the threadpool scheduler.

We have tests for `getInitialDelay` and to ensure a ScheduledFuture object is created. The only missing piece is to validate `scheduleWithFixedDelay()` executes as expected. This is a java util so I think it is a must-have to write our own tests considering it may not be viable.

If anyone knows a way I am happy to add it.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: Actually the `LifecycleListener` I was using will not work. It is only called when the ClusterManagerService class is created/killed. It will not always trigger when the cluster manager node changes.

I have reverted back to scheduling a job on all nodes, but early return on all nodes except the cluster manager. This is a simpler approach since we don't need to worry about the cluster manager changing and it adds little overhead - 1 call to clusterService per node per day.

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
ansjcy: Since we are using the `threadPool.scheduler().scheduleWithFixedDelay` function now, I'm good with the current tests.

ansjcy (commented):
Looks good to me!

ansjcy (approved):

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
ansjcy: this ^ makes sense to me.

ansjcy:
Have we tested this logic
```
clusterService.localNode().getRoles().contains(DiscoveryNodeRole.CLUSTER_MANAGER_ROLE)) {
                            deleteExpiredTopNIndices();
                        }
```
On a multi node cluster? If not we should do it later.

dzane17:
> On a multi node cluster? If not we should do it later.

Not yet. I will log this as a todo

dzane17:
I tested on a 3 node domain. Turns out we need to use **elected** cluster manager instead of cluster manager role. I have updated to `isLocalNodeElectedClusterManager()`. 

These logs show all 3 nodes return `true` for `isClusterManagerNode()` but only 1 node is the elected cluster manager.
```
[2025-04-09T18:59:46,145][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] isLocalNodeElectedClusterManager=false
[2025-04-09T18:59:46,147][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-1] isLocalNodeElectedClusterManager=true
[2025-04-09T18:59:46,146][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-2] isLocalNodeElectedClusterManager=false
[2025-04-09T18:59:46,146][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-0] isClusterManagerNode=true
[2025-04-09T18:59:46,148][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-1] isClusterManagerNode=true
[2025-04-09T18:59:46,147][INFO ][o.o.p.i.c.s.QueryInsightsService] [integTest-2] isClusterManagerNode=true
```

ansjcy:
>  Turns out we need to use elected cluster manager instead of cluster manager role.

Interesting, I assume there would only be one elected cluster manager on a running cluster (one node/multiple nodes scenarios)?

dzane17:
Yes, there is only one elected cluster manager at a time. The `CLUSTER_MANAGER_ROLE` is assigned to all nodes that are eligible to be cluster manager.

Ref: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/node/DiscoveryNodeRole.java#L244-L247

Issue #292, opened 2025-03-31
[BUG] Top_queries api illegal_argument_exception when `top_queries.latency.enabled` is false


dzane17:
### What is the bug?
When `search.insights.top_queries.latency.enabled` is false, top queries api (`GET /_insights/top_queries`) returns `illegal_argument_exception`. Even when top query cpu & memory are enabled. The api defaults to `type=latency` when no `type` is provided.

### How can one reproduce the bug?
1. Disable Top N by latency
```
% curl -X PUT "localhost:9200/_cluster/settings?pretty&flat_settings" -H 'Content-Type: application/json' -d '
{
  "transient" : {
    "search.insights.top_queries.latency.enabled" : false
  }
}'
{
  "acknowledged" : true,
  "persistent" : { },
  "transient" : {
    "search.insights.top_queries.latency.enabled" : "false"
  }
}
```
2. Call `top_queries` endpoint. Request will fail when no `type` param is provided.
```
% curl -XGET "http://localhost:9200/_insights/top_queries?from=2024-08-25T15:00:00.000Z&to=2024-08-30T17:00:00.000Z&pretty"
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "Cannot get top n queries for [latency] when it is not enabled."
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "Cannot get top n queries for [latency] when it is not enabled."
  },
  "status" : 400
}
```

### Solution
Do not check if metric type is enabled here - https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java#L358

Temporary work-around is to specify `type=cpu` or `type=memory` when CPU or memory is enabled, but latency is off.
```
// Valid if top queries CPU is enabled
GET /_insights/top_queries?type=cpu

// Returns Error
GET /_insights/top_queries
```

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

kkhatua:
@dzane17 Is this unique only to disabling latency, or would apply to any of the metrics being disabled, including 2 metrics being disabled ?

dzane17:
@kkhatua It is unique to latency because we default to `type=latency` when no type is present in the request.

[...]

PR #293, opened 2025-04-01
Upgrades to top_queries endpoint

dzane17:
### Description
This PR includes bug fixes for the Query Insights `top_queries` API.

1. Do not throw illegal argument exception from `top_queries` api when a metric type is disabled. The user may be trying to fetch historical data or data for other metric types.
2. We are currently limiting the number of records returned by the top_queries api to `top_n_size`. Note, this is different than the LocalIndexReader limit of 50. Removed this limit

### Testing
1. Disable top_n latency
```
% curl -X PUT "localhost:9200/_cluster/settings?pretty&flat_settings" -H 'Content-Type: application/json' -d '
{
  "transient" : {
    "search.insights.top_queries.latency.enabled" : false
  }
}'
{
  "acknowledged" : true,
  "persistent" : { },
  "transient" : {
    "search.insights.top_queries.latency.enabled" : "false"
  }
}
```
2. Call top queries API. Verify no error thrown
```
% curl -XGET "http://localhost:9200/_insights/top_queries?pretty"              
{
  "top_queries" : [ ]
}
```

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/292

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
@@ -306,7 +306,7 @@ public class TopQueriesService {
         @Nullable final String to,
         @Nullable final String id,
         @Nullable final Boolean verbose
-    ) throws IllegalArgumentException {
+    ) {
         OperationalMetricsCounter.getInstance()
             .incrementCounter(
                 OperationalMetric.TOP_N_QUERIES_USAGE_COUNT,
@@ -314,11 +314,6 @@ public class TopQueriesService {
                     .addTag(METRIC_TYPE_TAG, this.metricType.name())
                     .addTag(GROUPBY_TAG, this.queryGrouper.getGroupingType().name())
             );
-        if (!enabled) {
-            throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "Cannot get top n queries for [%s] when it is not enabled.", metricType.toString())
-            );
-        }
         // read from window snapshots
         final List<SearchQueryRecord> queries = new ArrayList<>(topQueriesCurrentSnapshot.get());
         if (includeLastWindow) {
@@ -363,14 +358,12 @@ public class TopQueriesService {
      * @return List of the records that are in local index (if enabled) with timestamps between from and to
      * @throws IllegalArgumentException if query insights is disabled in the cluster
      */
-    public List<SearchQueryRecord> getTopQueriesRecordsFromIndex(final String from, final String to, final String id, final Boolean verbose)
-        throws IllegalArgumentException {
-        if (!enabled) {
-            throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "Cannot get top n queries for [%s] when it is not enabled.", metricType.toString())
-            );
-        }
-
+    public List<SearchQueryRecord> getTopQueriesRecordsFromIndex(
+        final String from,
+        final String to,
+        final String id,
+        final Boolean verbose
+    ) {
         final List<SearchQueryRecord> queries = new ArrayList<>();
         final QueryInsightsReader reader = queryInsightsReaderFactory.getReader(TOP_QUERIES_READER_ID);
         if (reader != null) {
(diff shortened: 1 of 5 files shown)


src/test/java/org/opensearch/plugin/insights/QueryInsightsRestTestCase.java
```diff
@@ -354,14 +354,14 @@ protected String getTopQueries(String type) throws IOException {
...
                     break;
                 case "all":
-                    // Keep the default endpoint (no type parameter)
+                    // Keep the default endpoint (no sort parameter)
                     break;
                 default:
-                    // Throw an exception if the type is invalid
-                    throw new IllegalArgumentException("Invalid type: " + type + ". Valid types are 'all', 'cpu', 'memory', or 'latency'.");
+                    // Throw an exception if the sort is invalid
+                    throw new IllegalArgumentException("Invalid sort: " + type + ". Valid types are 'cpu', 'memory', or 'latency'.");
```
deshsidd, line 364: nit: Should ideally change the name of the `type` variable too?

src/test/java/org/opensearch/plugin/insights/rules/resthandler/top_queries/RestTopQueriesActionTests.java
deshsidd, line 1: Did we add a test to actually test the sorting behavior?

src/main/java/org/opensearch/plugin/insights/rules/transport/top_queries/TransportTopQueriesAction.java
```diff
@@ -116,10 +122,17 @@ protected TopQueries nodeOperation(final NodeRequest nodeRequest) {
         final String from = request.getFrom();
         final String to = request.getTo();
         final String id = request.getId();
-        return new TopQueries(
-            clusterService.localNode(),
-            queryInsightsService.getTopQueriesService(request.getMetricType()).getTopQueriesRecords(true, from, to, id)
+        Set<SearchQueryRecord> sortedSearchQueryRecordsSet = new TreeSet<>(
```
deshsidd, line 125: Could also make SearchQueryRecord implement Comparable?

ansjcy (commented):
We'll need to make some frontend changes to match this backend work. Like we need to update the details page to send API requests for all three metrics instead of just latency, and switch from type to sort.
Let's make sure we finish up the frontend PR so we can merge both the frontend and backend changes together. That way the query insights dashboards won't break in mainline/pipeline when the backend changes goes in.

ansjcy (changes requested):

src/main/java/org/opensearch/plugin/insights/rules/transport/top_queries/TransportTopQueriesAction.java
```diff
@@ -116,10 +111,17 @@ protected TopQueries nodeOperation(final NodeRequest nodeRequest) {
...
+            Comparator.comparingLong(record -> record.getMeasurement(request.getMetricType()).longValue())
         );
+        for (MetricType metricType : MetricType.values()) {
+            List<SearchQueryRecord> records = queryInsightsService.getTopQueriesService(metricType)
```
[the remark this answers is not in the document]
ansjcy: The in-memory priority queue already keeps the records of the highest values for each metric, if it is sort by, why do we need to add metrics from other queues? They will be either:
- irelavant (say if sort by latency, the latency queues now stores queries with >1000ms, the memory queues are all queres with < 10ms but high memory usage - this probably won't happen but can be an edge case - in this case adding queries from memory queue when sorting by latency won't make sense).
- Duplicated (Say the same set of queries are having high latency and high menory usage, we are basically adding the same queries twice to the final results.)

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -370,9 +356,7 @@ public List<SearchQueryRecord> getTopQueriesRecordsFromIndex(final String from,
                 List<SearchQueryRecord> records = reader.read(from, to, id);
                 Predicate<SearchQueryRecord> timeFilter = element -> start.toInstant().toEpochMilli() <= element.getTimestamp()
                     && element.getTimestamp() <= end.toInstant().toEpochMilli();
-                List<SearchQueryRecord> filteredRecords = records.stream()
-                    .filter(checkIfInternal.and(timeFilter))
-                    .collect(Collectors.toList());
+                List<SearchQueryRecord> filteredRecords = records.stream().filter(checkIfInternal.and(timeFilter)).toList();
```
dzane17, line 359: Use `.collect(Collectors.toList())` notation for compatibility with 2.x java versions

dzane17:
@deshsidd @ansjcy Reduced the scope of this PR after some discussion. No longer requires dashboard change. Please take a look. Thanks

ansjcy (approved):

Issue #285, opened 2025-03-25
[RFC] Tracking inflight top n queries in Query Insights


ansjcy:
### Is your feature request related to a problem?
Currently, the Top N Queries feature in Query Insights only tracks queries after they complete execution, whether successfully or unsuccessfully. As implemented in the [QueryInsightsListener.java ](https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java), queries are only added to the Top N queue once they finish processing. This approach, while valuable for historical analysis, creates an observability gap for long-running queries that are still in progress.

This RFC proposes extending the Top N Queries framework to track in-flight queries in real-time. This enhancement would allow users to understand the real-time impact of resource-intensive queries and potentially take proactive measures such as query cancellation to maintain system stability and resiliency.

### What solution would you like?
To implement support for tracking in-flight top N queries, we propose leveraging OpenSearch's existing Task Info API to collect real-time information about ongoing query operations and incorporate them into the Query Insights framework. Below are the important components/changes involved.

- Leveraging the Task API: We will utilize the GET Task API ([TransportGetTaskAction.java ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/cluster/node/tasks/get/TransportGetTaskAction.java)) to fetch information about currently running tasks across all nodes in the cluster, see [TaskInfo](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/tasks/TaskInfo.java#L73) for more details about the output, one example output is also attached at the end of the RFC.

- Task Correlation Mechanism: We'll develop a correlation algorithm that:
  - Get coordinator node search tasks (i.e. `SearchTask`, indices:data/read/search)
  - Correlates these with their corresponding data node tasks (i.e. `SearchShardTask`, indices:data/read/search[phase/query]) using the parent-child task relationship.
  - Extracts relevant metadata including query source, running time, and resource usage statistics, similar to how we are currently doing in the [QueryInsightsListener](https://github.com/opensearch-project/query-insights/blob/main/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java#L223).

[...]

PR #295, opened 2025-04-10
Inflight Queries API

ansjcy:
### Description
Add Inflight Queries API to fetch the ongoing search queries.
- Fetch running SearchTasks from nodes on demand and construct SearchQueryRecord from it.
- Extensive Unit tests coverage
- Integration test coverage for the API.

### Test
- Run an OpenSearch cluster with 2 nodes.
- Create document and perform 2 search requests on 2 nodes. Using Thread.sleep to simulate long running queries
- Get inflight queries:
```
➜  ~ curl -X GET "localhost:9200/_insights/live_queries?pretty"
{
  "live_queries" : [
    {
      "timestamp" : 1745359226777,
      "id" : "troGHNGUShqDj3wK_K5ZIw:512",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"userId\",\"boost\":1.0}}}}]",
      "node_id" : "troGHNGUShqDj3wK_K5ZIw",
      "measurements" : {
        "latency" : {
          "number" : 13959364458,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 3104,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 405000,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    },
    {
      "timestamp" : 1745359229158,
      "id" : "Y6eBnbdISPO6XaVfxCBRgg:454",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"userId\",\"boost\":1.0}}}}]",
      "node_id" : "Y6eBnbdISPO6XaVfxCBRgg",
      "measurements" : {
        "latency" : {
          "number" : 11579097209,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 3104,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 511000,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}
```

- Get in flight queries on specific node
```
curl -X GET "localhost:9200/_insights/live_queries?nodeId=integTest-0&pretty"
{
  "live_queries" : [
    {
      "timestamp" : 1745019842751,
      "id" : "faQoGlu2T8iAcJtIP3vSDQ:303",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"cyji\",\"boost\":1.0}}}}]",
      "node_id" : "faQoGlu2T8iAcJtIP3vSDQ",
      "measurements" : {
        "memory" : {
          "number" : 29560,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "latency" : {
          "number" : 7924553291,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 1971000,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}
```

- sort by
```
curl -X GET "localhost:9200/_insights/live_queries?sort=cpu&pretty"
{
  "live_queries" : [
    {
      "timestamp" : 1745358380622,
      "id" : "RrMW2BWRRfyrWQ47lPoGWg:89",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"userId\",\"boost\":1.0}}}}]",
      "node_id" : "RrMW2BWRRfyrWQ47lPoGWg",
      "measurements" : {
        "latency" : {
          "number" : 11261929250,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 4862000,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 547072,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    },
    {
      "timestamp" : 1745358378450,
      "id" : "fits7pNATQ6ZI6dbLJfn4Q:74",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"userId\",\"boost\":1.0}}}}]",
      "node_id" : "fits7pNATQ6ZI6dbLJfn4Q",
      "measurements" : {
        "latency" : {
          "number" : 13432647416,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 2359000,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 614720,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}
```

- size
```
curl -X GET "localhost:9200/_insights/live_queries?sort=cpu&size=1&pretty"
{
  "live_queries" : [
    {
      "timestamp" : 1745358437352,
      "id" : "fits7pNATQ6ZI6dbLJfn4Q:152",
      "description" : "indices[my-index-*], search_type[QUERY_THEN_FETCH], source[{\"size\":20,\"query\":{\"term\":{\"user.id\":{\"value\":\"userId\",\"boost\":1.0}}}}]",
      "node_id" : "fits7pNATQ6ZI6dbLJfn4Q",
      "measurements" : {
        "latency" : {
          "number" : 10969900667,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 394000,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 3168,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}
```

- After search query finishes
```
curl -X GET "localhost:9200/_insights/live_queries?nodeId=integTest-0&pretty"
{
  "live_queries" : [ ]
}
```

### Issues Resolved
Related to https://github.com/opensearch-project/query-insights/issues/285


dzane17:
@ansjcy Looks good. Do we need any of these features or are they planned for the future?

1. Sort – Add an option to sort by timestamp, latency, CPU, memory, node id. Or are we planning to handle this entirely in the dashboard?
2. Size/Limit – Limit the number of results returned
3. Duration Threshold – Filter for tasks running longer than a specified duration

ansjcy:
@dzane17 Those are all good points, by default I'm returning records sorted by latency.
We should do the sort/filter on backend. For now since I'm only considering the SearchTask, the resource usages won't make a lot of sense. But you are right, let me at least add those API parameters.

src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesResponse.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param params serialization parameters
+     * @param results live queries results from all nodes
+     * @throws IOException if an error occurs
+     */
+    private void toClusterLevelResult(final XContentBuilder builder, final Params params, final List<LiveQueries> results)
+        throws IOException {
+        final List<SearchQueryRecord> allQueries = results.stream()
+            .map(LiveQueries::getLiveQueries)
+            .flatMap(Collection::stream)
+            .sorted((a, b) -> SearchQueryRecord.compare(a, b, MetricType.LATENCY) * -1)
```
deshsidd, line 89: Are we planning to add sort param in this PR?

deshsidd: Also what are your thoughts about including size and duration as suggestion by @dzane17 ?

src/main/java/org/opensearch/plugin/insights/rules/transport/live_queries/TransportLiveQueriesAction.java
```diff
@@ -0,0 +1,144 @@
...
+
+        // Execute tasks request asynchronously to avoid blocking transport threads
+        client.admin().cluster().listTasks(listTasksRequest, new ActionListener<ListTasksResponse>() {
+            @Override
+            public void onResponse(ListTasksResponse taskResponse) {
+                try {
+                    // Group tasks by node and build records
+                    Map<String, List<SearchQueryRecord>> perNode = new HashMap<>();
+                    for (TaskInfo taskInfo : taskResponse.getTasks()) {
+                        if (!taskInfo.getAction().startsWith("indices:data/read/search")) {
```
deshsidd, line 84: We could also capture the current phase/type of search in the future?

```
indices:data/read/search[phase/query]

indices:data/read/search[phase/fetch]

indices:data/read/search/shard

indices:data/read/search/scroll

indices:data/read/search/template
```

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -80,6 +80,12 @@ public class QueryInsightsSettings {
      *
      */
     public static final String TOP_QUERIES_BASE_URI = PLUGINS_BASE_URI + "/top_queries";
+
+    /**
+     * Settings for Live Queries
+     */
+    public static final String LIVE_QUERIES_BASE_URI = PLUGINS_BASE_URI + "/live_queries";
```
deshsidd, line 87: No settings needed for this to enable/disable this feature?

src/test/java/org/opensearch/plugin/insights/rules/resthandler/live_queries/LiveQueriesRestIT.java
```diff
@@ -0,0 +1,371 @@
...
+
+                    // Validate aggregationType is a string
+                    assertTrue("AggregationType should be a string", metric.get("aggregationType") instanceof String);
+                }
+                assertTrue("Description should be a string", query.get("description") instanceof String);
+            }
+        } else {
+            logger.info(
+                "No live queries found after {} attempts - this is acceptable due to test environment limitations",
+                MAX_POLL_ATTEMPTS
```
deshsidd, line 218: Hope this does not cause the tests to silently pass in scenarios where it should fail ? Eg: cases where some code regressions cause the live queries api to return empty each time.

ansjcy:
There are issues with the current implementation: we are fetching task level information from coordinator level, thus we don't need Node level request and response objects - we can consider adding it later because we might need to correlate shard level information, but doing it now seems over engineering. I'll refactor this now.

src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesResponse.java
ansjcy: Yes as I mentioned here: https://github.com/opensearch-project/query-insights/pull/295#issuecomment-2816394705

src/main/java/org/opensearch/plugin/insights/rules/transport/live_queries/TransportLiveQueriesAction.java
ansjcy: No, the purpose of this PR is to add coordinator level search task only. All tasks you mentioned should belong to the data node or unrelated to search task. Even if we add it here, we won't be able to correlate to the actual search task.

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
ansjcy: No, this would be an on-demand API. I checked with PM/UX, there's no need to add settings for on-demand APIs.

src/test/java/org/opensearch/plugin/insights/rules/resthandler/live_queries/LiveQueriesRestIT.java
ansjcy: yeah potentially it will. The actual issue is we won't be able to simulate slow queries in a reliable way. I can try commenting out/ removing this part for now and see if this test would be flaky though.

ansjcy: Just updated the PR and removed this fallback. I also ran these tests without this fallback for 100 times and it is always passing. So my previous assumption that this case can be flaky is wrong.
```
TOTAL_RUNS=100
failures=0

for ((i=1; i<=TOTAL_RUNS; i++)); do
  echo "=== Run $i/$TOTAL_RUNS ==="
  if ! ./gradlew integTest --tests "org.opensearch.plugin.insights.rules.resthandler.live_queries.LiveQueriesRestIT"; then
    echo "Test failed on run $i"
    ((failures++))
  fi
  echo
done
successes=$((TOTAL_RUNS - failures))
echo "=== Summary ==="
echo "Total runs   : $TOTAL_RUNS"
echo "Successes    : $successes"
echo "Failures     : $failures"
```
Results:
```
=== Summary ===
Total runs   : 100
Successes    : 100
Failures     : 0
```

ansjcy:
Updated to address the issues mentioned in my above comments and added parameters
https://github.com/opensearch-project/query-insights/pull/295#issuecomment-2819762552

I'm adding `sort`, `size`, `nodeId` for now. I leave the threshold for now because it can be a little bit tricky when used together with sort and size. I think we need further discussions on `threshold` and cannot be done in the 3.0 timeframe. cc @dzane17

src/main/java/org/opensearch/plugin/insights/rules/transport/live_queries/TransportLiveQueriesAction.java
```diff
@@ -0,0 +1,131 @@
...
+                            taskInfo.getTaskId().toString()
+                        );
+                        allFilteredRecords.add(record);
+                    }
+
+                    // Sort descending by the requested metric and apply size limit in one pass
+                    List<SearchQueryRecord> finalRecords = allFilteredRecords.stream()
+                        .sorted((a, b) -> SearchQueryRecord.compare(b, a, request.getSortBy()))
+                        .limit(request.getSize() < 0 ? Long.MAX_VALUE : request.getSize())
+                        .toList();
```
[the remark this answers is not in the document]
ansjcy: I don't think we want to backport the 3.0 features to 2.x anymore since they will be in maintenance mode, so should this be fine?

src/main/java/org/opensearch/plugin/insights/rules/resthandler/live_queries/RestLiveQueriesAction.java
```diff
@@ -0,0 +1,87 @@
...
+        final LiveQueriesRequest liveQueriesRequest = prepareRequest(request);
+
+        return channel -> client.execute(LiveQueriesAction.INSTANCE, liveQueriesRequest, liveQueriesResponse(channel));
+    }
+
+    static LiveQueriesRequest prepareRequest(final RestRequest request) {
+        final String[] nodesIds = Strings.splitStringByCommaToArray(request.param("nodeId"));
+        final boolean verbose = request.paramAsBoolean("verbose", true);
+        final String sortParam = request.param("sort", MetricType.LATENCY.toString());
+        final MetricType sortBy = MetricType.fromString(sortParam);
```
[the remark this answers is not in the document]
ansjcy: Good point, now the error message is like
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "No enum constant org.opensearch.plugin.insights.rules.model.MetricType.ANY"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "No enum constant org.opensearch.plugin.insights.rules.model.MetricType.ANY"
  },
  "status" : 400
}
```
Let me refactor this part.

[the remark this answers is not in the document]
ansjcy: Added validation.

dzane17 (approved):

Issue #251, opened 2025-02-27
[FEATURE] Incorporate indexed priority query in query grouping logic


deshsidd:
### Is your feature request related to a problem?
_A clear and concise description of what the problem is, e.g. I'm always frustrated when [...]._
Currently we use a priority queue, incorporate indexed priority queue in the grouping logic to make the time complexity go from O(n) to O(logn). 

### What solution would you like?
_A clear and concise description of what you want to happen._

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

Dilaxn:
Is your feature request related to a problem?
Currently, the grouping logic uses a standard priority queue, causing the operation to have O(n) complexity in some cases. This becomes inefficient and leads to performance bottlenecks, especially for larger data sets.

What solution would you like?
Introduce an Indexed Priority Queue in the query grouping logic.
This change reduces time complexity from O(n) to O(log n), significantly improving performance during frequent updates.

What alternatives have you considered?
Binary Heap without indexing: Simple, but updates would still require linear scans (O(n)).
Fibonacci Heap or Pairing Heap: Could reduce complexity further, but are more complex and less practical for the specific case.

Do you have any additional context?
I've reviewed the current implementation and believe that integrating an Indexed Priority Queue is practical and effective. I'll ensure the new implementation includes sufficient test coverage and performance benchmarks.


Note : If I get support it will be more helpful. Thank you.

PR #299, opened 2025-04-13
Implement Indexed Priority Queue to improve grouping performance (#251)

Dilaxn:
Is your feature request related to a problem?
Yes. The current grouping logic uses a standard priority queue, forcing removal operations to run in O(n)O(n). This creates a significant performance bottleneck—especially in scenarios with frequent updates or large data sets.

What solution would you like?
Introduce an Indexed Priority Queue in the query grouping logic. By allowing direct lookup of an element’s position in the heap, arbitrary removals and updates can be done in O(log⁡n) time instead of O(n). This significantly boosts performance for frequent insertions and removals.

What alternatives have you considered?

1.Binary Heap without indexing: Simpler to implement, but removal of an arbitrary element still requires an O(n) scan.

2.Fibonacci Heap or Pairing Heap: While they can provide even better theoretical performance for certain operations, they are more complex and less practical for our specific use case.

Additional context
I have reviewed the existing code and tested how the new Indexed Priority Queue integrates with the grouping logic. Early benchmarks show a noticeable reduction in latency for frequent queue updates. I plan to add thorough test coverage to validate correctness and performance improvements before merging.


kkhatua:
@Dilaxn Thank you for taking interest in improving performance. Unfortunately we don't have any performance tests as yet to validate this. Would you be able to help share some numbers so as to showcase the performance gains?  (cc: @deshsidd  )

src/main/java/org/opensearch/plugin/insights/core/utils/IndexedPriorityQueue.java
```diff
@@ -0,0 +1,127 @@
...
+    private final Map<T, Integer> indexMap;
+    private final Comparator<? super T> comparator;
+
+    public IndexedPriorityQueue(int initialCapacity, Comparator<? super T> comparator) {
+        this.comparator = comparator;
+        this.heap = new ArrayList<>(initialCapacity);
+        this.indexMap = new HashMap<>(initialCapacity);
+    }
+
+    public synchronized boolean insert(T item) {
```
ansjcy, line 30: As I think about it more, using `synchronized` on each function is a good start to ensure thread safety, but it might be a bottleneck under high contention. I think it makes more sense to use more fine-grained locking mechanisms instead of doing `synchronized` for all public functions - the main challenge here should be ensuring data consistency between the heap and the indexMap (e.g. in `siftUp`, updating on the heap and indexMap should be atomic).

ansjcy (commented):
@Dilaxn Thanks for your work on this! I've added several comments regarding the generic type handling and thread safety.

@deshsidd, could you please also take a look to verify the functional correctness of the Indexed Priority Queue logic? Thanks!

ansjcy:
Also the unit test `testAddRecordToLimitAndDrain` is failing, could you double check? 
https://github.com/opensearch-project/query-insights/actions/runs/14431905665/job/40541189714?pr=299

You can run `./gradlew test` on your local env to validate your changes as well.

Dilaxn:
Thank you @ansjcy  and @kkhatua for your feedback. I do check the review comments and try to solve those. Thank you.

src/main/java/org/opensearch/plugin/insights/core/service/TopQueriesService.java
```diff
@@ -99,7 +99,7 @@ public class TopQueriesService {
     /**
      * The internal thread-safe store that holds the top n queries insight data
      */
-    private final PriorityBlockingQueue<SearchQueryRecord> topQueriesStore;
+    private final IndexedPriorityQueue<SearchQueryRecord> topQueriesStore;
```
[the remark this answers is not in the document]
Dilaxn: We’re already handing the very same topQueriesStore into the MinMaxHeapQueryGrouper, so splitting it into two separate queues just adds needless complexity. By switching both service and grouper to use a single IndexedPriorityQueue, we get one shared heap—no extra merge logic, no change in performance characteristics—and all of our inserts, polls, and snapshots happen against the exact same data structure. For those reasons, I’d stick with the IPQ here.

src/main/java/org/opensearch/plugin/insights/core/utils/IndexedPriorityQueue.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class IndexedPriorityQueue<T> {
+
+    private final List<T> heap;
+    private final Map<T, Integer> indexMap;
```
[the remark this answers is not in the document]
Dilaxn: Keys are now of type K, separate from your mutable V values, so you only ever hash or compare on that stable identifier.

Dilaxn:
Hi @ansjcy I’ve updated the code per your review:
-Replaced T with Entry<K, V>.
-Refined the locking mechanism.
-Applied fixes to the unit tests.
Please verify the changes, Thank you for ur time and consideration.

src/main/java/org/opensearch/plugin/insights/core/utils/IndexedPriorityQueue.java
```diff
@@ -0,0 +1,185 @@
...
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+public class IndexedPriorityQueue<K, V> {
```
deshsidd, line 19: Thanks for the changes! We should ideally use IndexedPriorityQueue from a stable source rather than trying to implement it ourselves?

This implementation seems the most popular one used in industry: https://algs4.cs.princeton.edu/24pq/

deshsidd (changes requested):

ansjcy:
As .kkhatua mentioned:
> Would you be able to help share some numbers so as to showcase the performance gains?

We might need to do some benchmarking tests and CPU profiling to understand the performance gains. We need to:
- run an OpenSearch cluster
- use https://docs.opensearch.org/docs/latest/benchmark/ to ingest some workload. 
- Then run some search requests with different shapes. 
- use https://github.com/async-profiler/async-profiler to see the CPU usage impact numbers.

deshsidd:
@Dilaxn Hope you are able to make progress on this PR. Please let us know if you need any guidance :)

Dilaxn:
@deshsidd @ansjcy 
Sorry for the delay. I’ve been testing the functionality and performance more thoroughly, which is why it took some extra time. I’ll be working on it this Saturday and Sunday and will update the GitHub issue accordingly.
Thanks a lot for your support and understanding!

Dilaxn:
@ansjcy @kkhatua 
[PQ.json](https://github.com/user-attachments/files/20067063/PQ.json)
[IPQ.json](https://github.com/user-attachments/files/20067064/IPQ.json)
Attached the benchmark results comparing the PriorityBlockingQueue (PQ) version of the Query Insights plugin with the updated IndexedPriorityQueue (IPQ) version. Here's a summary of key performance metrics:
| Metric                               | PQ            | IPQ               | Δ            |
| ------------------------------------ | ------------- | ----------------- | ------------ |
| **Index throughput (mean)**          | 50,856 docs/s | **61,300 docs/s** | **↑ 20.5%** |
| p90 latency                          | 1,203 ms      | **804 ms**        | ↓ 33%       |
| p99 latency                          | 5,929 ms      | **2,043 ms**      | ↓ 65%       |
| **Indexing duration**                | 3,532 s       | **2,825 s**       | ↓ 20%       |
| **Distance agg service time (mean)** | 6,250 ms      | **11,173 ms**     | **↑ 78%**   |
| Merge time                           | 6,422 s       | **5,886 s**       | ↓ 8%        |
| Segment count                        | 27            | **21**            | ↓ 22%       |
| Young GC time                        | 167 s         | **145 s**         | ↓ 13%       |


Key Takeaways:

    The IPQ-based version significantly improves indexing throughput and reduces latency, especially at p99.

    A slight increase in distance aggregation service time was observed 
    Overall, the switch to IPQ results in lower GC pressure, faster merging, and fewer segments.

Please feel free to correct me if there’s any mistake. Thank you!

Dilaxn:
Subject: ✅ Functional Parity & Concurrency Benchmark Validation for IPQ

Hi @deshsidd ,

I've completed a comprehensive validation of our IndexedPriorityQueue (IPQ) against the suggested IndexMinPQ implementation.
🔍 Functional Parity

    ✅ Verified 1:1 behavioral parity using:

        1,000 randomized functional test scenarios

        Property-based tests (250+ diverse operation sequences)

        Edge case coverage (insert, remove, peek, poll correctness)

    🎯 Both queues produced identical sorted outputs and consistent state transitions.

⚙️ Concurrency Comparison

    Benchmarked concurrent insertions (8 threads, 100,000 entries) against sequential IndexMinPQ.

    Observation: IPQ performs slightly slower than the non-thread-safe version in microbenchmarks, primarily due to the overhead introduced by ReentrantReadWriteLock — which is expected for thread-safe implementations.

    Advantage: IPQ is concurrency-safe by design and scales better under real-world multithreaded conditions (read/write workloads).

For this, I created a separate Gradle project with the help of GPT. The project is :  https://github.com/Dilaxn/indexed-priority-queue-compat

Thanks again for the guidance!

Best regards,
Dilakshan

Dilaxn:
> @Dilaxn This is very impressive, thanks for doing the performance benchmarking! Let me take a final check on the PR. Also, since this change mainly affects search path, I'm wondering why are we seeing an improvement on indexing. Do you have more performance numbers on the search path?

Hi @ansjcy,  Thanks for the feedback!
I’ve attached the benchmark reports for the nyc_taxis workload using both IndexedPriorityQueue (the new change) and PriorityBlockingQueue (master). 
[IndexedPriorityQueue.json](https://github.com/user-attachments/files/20638307/IndexedPriorityQueue.json)
[master benchmark.txt](https://github.com/user-attachments/files/20638309/master.benchmark.txt)
[PriorityBlockingQueue.json](https://github.com/user-attachments/files/20638310/PriorityBlockingQueue.json)

deshsidd:
@Dilaxn Thanks for the parity checks and the benchmarks. This is pretty impressive! Let me take a final look today.

src/main/java/org/opensearch/plugin/insights/core/utils/IndexedPriorityQueue.java
deshsidd: Discussed offline. Seems like the princeton implementation might not be thread safe. Going with this approach for now.

deshsidd, line 19: Consider logging when operations like insert() or remove() fail due to key presence/absence, especially in debug mode?

src/main/java/org/opensearch/plugin/insights/core/utils/IndexedPriorityQueue.java
```diff
@@ -0,0 +1,185 @@
...
+        try {
+            Integer idx = indexMap.remove(key);
+            if (idx == null) return false;
+
+            int lastIdx = heap.size() - 1;
+            if (idx != lastIdx) {
+                Entry<K, V> lastItem = heap.get(lastIdx);
+                heap.set(idx, lastItem);
+                indexMap.put(lastItem.key, idx);
+                siftDown(idx);
```
deshsidd, line 70: Can we avoid calling both siftDown and siftUp in this case by doing some checks?

deshsidd (commented):
LGTM overall. Thanks again for the change. Some minor comments @Dilaxn 
Once addressed we can aim to rebase approve and merge.

ansjcy:
hmm, looking at the full benchmark run results I see latency increases on indexed PQ on almost all search workloads. I want us to do some extra benchmarks and cpu profilings. Let's hold on this PR for now until we fully understand the performance impact.

ansjcy:
@Dilaxn has done some very good experiments on the performance of indexed priority queue but based on the performance numbers we need to re-evaluate the implementation. It could be the implementation itself or the native java pq implementation has more optimizations that we have not added to this PR. This is not a trival problem to solve, I'm closing this PR for now for future references in case we need to revisit in the future.

Issue #354, opened 2025-06-06
[FEATURE] Add isCancelled field to Live Queries API response


KishoreKicha14:
### Is your feature request related to a problem?

The current Live Queries API does not expose whether a query has been cancelled or not. This enhancement proposes adding an isCancelled boolean field to each live query object returned by the API. This will help front-end applications like the Query Insights Dashboard clearly indicate and manage the state of running and cancelled queries.

### What solution would you like?

Update the Live Queries response schema to include a new field: 

`"isCancelled": true | false`

### Do you have any additional context?

Enhances observability and control in dashboards and monitoring tools

PR #355, opened 2025-06-06
Added isCancelled field in Live Queries API

KishoreKicha14:
### Description

This PR introduces an isCancelled boolean field in the Live Queries API response to indicate whether a query has been cancelled.

### Changes Introduced:

- Added isCancelled field to the LiveSearchQueryResponse model.
- Updated LiveQueriesTransportAction to populate the isCancelled value.

Testing Done:

Verified API responses include the isCancelled field with correct values.

### Example response

```
{
  "ok": true,
  "response": {
    "live_queries": [
      {
        "timestamp": 1749187466964,
        "id": "node-A1B2C3D4E5:3600",
        "description": "indices[top_queries-2025.06.06-11009], search_type[QUERY_THEN_FETCH]",
        "node_id": "node-A1B2C3D4E5",
        "measurements": {
          "latency": {
            "number": 7990852130,
            "count": 1,
            "aggregationType": "NONE"
          },
          "cpu": {
            "number": 89951,
            "count": 1,
            "aggregationType": "NONE"
          },
          "memory": {
            "number": 3818,
            "count": 1,
            "aggregationType": "NONE"
          },
          "is_cancelled": true
        }
      }
}
```

### Issues Resolved
Closes [#354]

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
@@ -47,6 +47,8 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
     private final Map<MetricType, Measurement> measurements;
     private final Map<Attribute, Object> attributes;
     private final String id;
+    private final boolean isCancelled;
+
 
     /**
      * Timestamp
@@ -154,8 +156,8 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
             });
         }
         this.attributes = Attribute.readAttributeMap(in);
+        this.isCancelled = false;
     }
-
     /**
      * Constructor of SearchQueryRecord
      *
@@ -164,7 +166,7 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
      * @param attributes A list of Attributes associated with this query
      */
     public SearchQueryRecord(final long timestamp, Map<MetricType, Measurement> measurements, final Map<Attribute, Object> attributes) {
-        this(timestamp, measurements, attributes, UUID.randomUUID().toString());
+        this(timestamp, measurements, attributes, UUID.randomUUID().toString(), false);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -154,6 +155,7 @@ public SearchQueryRecord(final StreamInput in) throws IOException, ClassCastExce
             });
         }
         this.attributes = Attribute.readAttributeMap(in);
+        this.isCancelled = false;
```
dzane17, line 158: I think we need to read the boolean from the input stream and add a version check:
```
if (in.getVersion().onOrAfter(Version.V_3_1_0)) {
    this.isCancelled = in.readBoolean();
} else {
    this.isCancelled = false;
}
```
Also update the [writeTo](https://github.com/opensearch-project/query-insights/blob/bcae2c97c63e218789365312cca1bf8726ba3bdc/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java#L524) method in a similar fashion unless this info is not needed in node-to-node communication?

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -176,18 +178,17 @@ public SearchQueryRecord(final long timestamp, Map<MetricType, Measurement> meas
      * @param id unique id for a search query record
      */
     public SearchQueryRecord(
-        final long timestamp,
+        long timestamp,
         Map<MetricType, Measurement> measurements,
-        final Map<Attribute, Object> attributes,
```
dzane17, line 181: Why are `timestamp` and `attributes` non-final?

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -47,6 +47,7 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
     private final Map<MetricType, Measurement> measurements;
     private final Map<Attribute, Object> attributes;
     private final String id;
+    private final boolean isCancelled;
```
dzane17, line 50: Wondering if it's better to store `isCancelled` in the `attributes` Map? Then we don't need version checks in the streamInput constructor and writeTo method.

ansjcy: +1, let's add it into the attribute map.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -181,12 +185,9 @@ public SearchQueryRecord(
         final Map<Attribute, Object> attributes,
         String id
     ) {
-        if (measurements == null) {
```
ansjcy, line 184: why do we remove this

ansjcy:
There's a flaky test failure not related to this PR. Hopefully it can be resolved by https://github.com/opensearch-project/query-insights/pull/364. Merging it now

PR #365, opened 2025-06-11
Fix a bug in creating node level top queries request

ansjcy:
### Description
Fix a bug in creating node level top queries request and change default integ test node number to 2 (and fixed the failed integ tests because of this change) so that we can capture similar issues in the future.

The bug is the request parameters are not passed correctly to other nodes.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesRequest.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesRequest.java
@@ -9,6 +9,7 @@
 package org.opensearch.plugin.insights.rules.action.top_queries;
 
 import java.io.IOException;
+import org.opensearch.Version;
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesRequest.java
```diff
@@ -34,10 +34,10 @@ public class TopQueriesRequest extends BaseNodesRequest<TopQueriesRequest> {
...
         super(in);
         this.metricType = MetricType.readFromStream(in);
-        this.from = null;
-        this.to = null;
-        this.verbose = null;
-        this.id = null;
+        this.from = in.readOptionalString();
+        this.to = in.readOptionalString();
+        this.id = in.readOptionalString();
+        this.verbose = in.readOptionalBoolean();
```
dzane17, line 40: I think we need version checks for reading and writing these params. Older version nodes will not send or expect these.

PR #424, opened 2025-09-23
[Split into 3 PRs now] Optimize query storage by storing source field as a string in local index

LilyCaroline17:
### Update
Split into 3 PRs: #479, #483, #484

### Description
This PR addresses some bugs that we're having with how the source field is stored in the local index. The source field can be really long and/or really nested. The solution is to store the source field as a string within the local index and in-memory so no exceptions related to field depth or number of fields is triggered and to maintain consistency. 

For the top_queries index template that contains the local index mapping, the template is updated daily during index creation. When changing the index mapping, upgrades should not be affected because reindex automatically upgrades data format from object to string. 

Additionally, for strings that are too long, they are truncated and a boolean to mark that the string has been truncated is set. These source strings are then skipped when parsing. 

### Issues Resolved
Closes [opensearch-project/query-insights/issues/352] and [opensearch-project/query-insights/issues/411]
Dashboards PR to handle source field change: https://github.com/opensearch-project/query-insights-dashboards/pull/357

Check that Local index properly created: 

Screenshot 2025-09-29 at 11 32 01 AM

Scrolling down we can see that source is in properties in the local index. 

Screenshot 2025-09-29 at 11 39 43 AM

### Test Results (1000 query records)

I performed a test with 1000 query records to compare the storage usage of storing source as a string compared to storing source as a SearchSourceBuilder and got the following results: 

| Query Complexity | SearchSourceBuilder | String Storage | Memory Saved |
|------------------|-------------------|----------------|--------------|
| Small queries | 320B/record | 312B/record | **2.6%** |
| Average queries | 1.8KB/record | 751B/record | **58.4%** |
| Large queries | 4.0KB/record | 2.0KB/record | **49.3%** |
| Very large (truncated) | 10.3KB/record | 4.4KB/record | **57.4%** |

We can see that there is a significant reduction in memory storage for source when it's stored as string, especially in larger, more complex queries. 

However, it's important to note that source was stored as a SearchSourceBuilder for a reason, and after it's stored into local index, it's needed as a SearchSourceBuilder only in SearchQueryCategorizer, which only affects deployments with search query metrics enabled. This extra cost would be 4-15 ms per query parsing overhead. 

The memory saving benefits and stability (avoiding failures from attempts to store source as a SearchSourceBuilder) take precedence over the extra time it takes to parse queries in certain conditions. 


ansjcy:
We should do more tests especially for compatibilities in this PR, e.g.
- Run a cluster with existing top queries indices that have the old mapping
- enable export for different settings on the MAX_SOURCE_LENGTH, until truncating happens.
- You can include test results for retrieving top queries from local index for those different scenarios in the PR description.

PR #441, opened 2025-10-06
Use ARCHIVE test distribution for testing with testclusters

cwperks:
### Description

Use ARCHIVE test distribution for testing with testclusters

The `INTEG_TEST` distribution is a very minimal distribution for testing which only includes transport modules. The ARCHIVE distribution will use the latest min distro snapshot (at least that's my understanding - not sure how it actually translates to code).

### Issues Resolved

Fixes https://github.com/opensearch-project/query-insights/actions/runs/18180056333/job/52078406568?pr=439

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -318,7 +318,7 @@ tasks.named("integTest").configure {
 }
 
 testClusters.all {
-    testDistribution = "INTEG_TEST"
+    testDistribution = "ARCHIVE"
     // This installs our plugin into the testClusters
     plugin(project.tasks.bundlePlugin.archiveFile)
 


cwperks:
I finally learned the distinction between ARCHIVE and INTEG_TEST distribution for testclusters.

When using ARCHIVE it uses the min distro for the latest published snapshot (though not sure how to track that since its done off github)

When using INTEG_TEST it actually uses the :distribution:archive:integ-test-zip distribution

Interestingly, INTEG_TEST does not include all modules. It only includes modules matching [transport-](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/build.gradle#L233-L235). I have no idea why that's the case, but TIL

Issue #453, opened 2025-10-16
[BUG] Query Insights exporter delete the local saved indices if customer disables the exporter


bouhajer:
### What is the bug?
If you enable the exporter in Query insights, this will create a daily indices storing the top N queries. However, if you later disable the exporter, putting it to "None", this will delete all the historical indices previously created by the plugin.
### How can one reproduce the bug?
Enable the exported in Query insights and then disable it.

### What is the expected behavior?
Query insights exporter should not delete any historical indices created by the plugin before the end of retention period set by the customer. If needed, customers can manually delete the created indices.

### What is your host/environment?
OpenSearch 3.1

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
The customer has already set the retention period for their indices, so the plugin should honour that period and not delete any indices before the end of that period.

ansjcy:
Thanks for opening the issue. Currently this is the expected behavior but I agree it might be unexpected for users and they will potentially lose top queries data without warning. We should at least add a warning message around this.

Or, We can make disabling local index be non-destructive (preserve historical data), instead of deleting everything. we can add separate a "Delete Data" API. but in this case, our retention policies need to be independent of enable/disable state.

thoughts? @dzane17

PR #465, opened 2025-10-20
Retain local indices on exporter type change

dzane17:
### Description
- Remove logic that deletes all Query Insights indices when exporter type setting is changed
- Run daily deletion job regardless of current exporter
- Introduced `LocalIndexLifecycleManager` to hold all delete-after state and logic

### Issues Resolved
Resolves https://github.com/opensearch-project/query-insights/issues/453

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
@@ -10,11 +10,8 @@ package org.opensearch.plugin.insights.core.service;
 
 import static org.opensearch.plugin.insights.core.service.TopQueriesService.TOP_QUERIES_EXPORTER_ID;
 import static org.opensearch.plugin.insights.core.service.TopQueriesService.TOP_QUERIES_READER_ID;
-import static org.opensearch.plugin.insights.core.service.TopQueriesService.isTopQueriesIndex;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_GROUPING_TYPE;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_TOP_N_QUERIES_INDEX_PATTERN;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.MAX_DELETE_AFTER_VALUE;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.MIN_DELETE_AFTER_VALUE;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.QUERY_INSIGHTS_EXECUTOR;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_EXPORTER_DELETE_AFTER;
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_N_EXPORTER_TYPE;
@@ -22,15 +19,12 @@ import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_
 import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -704,26 +698,6 @@ void deleteExpiredTopNIndices() {
         }
     }
 
-    /**
-     * Deletes all Top N local indices
-     *
-     * @param client OpenSearch Client
-     * @param localIndexExporter the exporter to handle the local index operations
-     */
-    void deleteAllTopNIndices(final Client client, final LocalIndexExporter localIndexExporter) {
```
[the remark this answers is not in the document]
dzane17: No, admins can delete indices manually I believe and retention policy will still apply

ansjcy:
Let's also create a documentation PR for this, we missed adding the expectation for disabling the local index (i.e. warning: local index will be deleted after you disable local index). We should call this out now in the doc: your index will not be deleted, if you need to delete, do it manually.

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
```diff
@@ -324,61 +320,6 @@ public void testGetHealthStats() {
         assertTrue(fieldTypeCacheStats.containsKey(MISS_COUNT));
     }
 
-    public void testDeleteAllTopNIndices() throws IOException {
```
ansjcy: instead of removing the tests, we should have tests to validate that the local index "will not be deleted" after disabling.

ansjcy:
I have some deeper concerns on the retention policy behavior after this change.
1. If we disable local index exporter, that `delete_after` job will not delete the expired indices now because the related logic will only run if local index exporter is enabled https://github.com/opensearch-project/query-insights/blob/34391ac34311dfeadc4d7e526c561263d80bcdba/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java#L679, these indices will become orphaned forever.
2. If "enable local index" -> "disable all features" , because of this check https://github.com/opensearch-project/query-insights/blob/34391ac34311dfeadc4d7e526c561263d80bcdba/src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java#L604, the old indices will also remain there forever. 

Let's think about even more scenarios: like
- local_index -> disable all features -> reenable with local index
- local_index -> disable all -> reenable with none
...

IMO we should always enforce the retention job, regardless whether features are enabled, or what type of exporter is enabled. we can also consider adding "none" to delete after , so that people can also disable this retention job.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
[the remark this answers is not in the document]
dzane17: That is a good point. I added the ability to immediately delete all local indices by setting `delete_after_days = 0`. Previously the min value was 1. 0 means no retention period so I think it makes sense to use as a "delete now" option.

src/test/java/org/opensearch/plugin/insights/core/service/QueryInsightsServiceTests.java
dzane17: Updated and added tests

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -185,16 +175,16 @@ public QueryInsightsService(
...
                 TOP_N_EXPORTER_DELETE_AFTER,
-                (this::setExporterDeleteAfterAndDelete),
-                (this::validateExporterDeleteAfter)
+                (localIndexLifecycleManager::setDeleteAfterAndDelete),
+                (localIndexLifecycleManager::validateDeleteAfter)
             );
         clusterService.getClusterSettings()
             .addSettingsUpdateConsumer(TOP_N_EXPORTER_TEMPLATE_PRIORITY, this::setTemplatePriority, this::validateTemplatePriority);
 
-        this.setExporterDeleteAfterAndDelete(clusterService.getClusterSettings().get(TOP_N_EXPORTER_DELETE_AFTER));
```
ansjcy: where is the corresponding logic for this line in the life cycle manager? otherwise if we changed the cluster setting and the service is restarted, the LocalIndexLifecycleManager will always use the default value instead of the modified value.

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -610,21 +561,21 @@ protected void doStart() {
...
-                        }
-                    } catch (Exception e) {
-                        logger.error("Error occurred while deleting expired indices:", e);
+        }
+        if (threadPool.scheduler() != null) {
+            deleteIndicesScheduledFuture = threadPool.scheduler().scheduleWithFixedDelay(() -> {
+                try {
+                    if (clusterService.isStateInitialised()
+                        && clusterService.state().getNodes() != null
+                        && clusterService.state().getNodes().isLocalNodeElectedClusterManager()) {
```
ansjcy: should we skip this job if delete after is 0?

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
dzane17: That is a good catch. I will add `deleteAfter` as an arg in the `LocalIndexLifecycleManager` constructor and call `setDeleteAfterAndDelete()` inside the constructor. That way we don't ever need to call `setDeleteAfterAndDelete` separately. It will always have the fresh value on instantiation.

PR #479, opened 2025-11-10
Remove index template

LilyCaroline17:
### Description
Removed index template from LocalIndexExporter to solely rely on explicit index mappings to create new indexes. Mapping is always applied during index creation and takes precedence over index templates in  OpenSearch, meaning that as long as our mapping has everything we need in our index, it will be there for index creation. User index templates would still be applied for any fields not defined in our mapping, only filling in the gaps in fields that we do not use in query insights. 

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
@@ -9,18 +9,14 @@
 package org.opensearch.plugin.insights.core.exporter;
 
 import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_DELETE_AFTER_VALUE;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.DEFAULT_TEMPLATE_PRIORITY;
-import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.TOP_QUERIES_INDEX_PATTERN_GLOB;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.concurrent.CompletableFuture;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.ExceptionsHelper;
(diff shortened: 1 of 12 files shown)


ansjcy (commented):
I also see a drop on code coverage score.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -328,28 +331,61 @@ String readIndexMappings() throws IOException {
...
             @Override
             public void onResponse(GetComposableIndexTemplateAction.Response response) {
-                // If the template exists and priority has not been changed, we don't need to create/update it
-                if (response.indexTemplates().containsKey(TEMPLATE_NAME)
-                    && response.indexTemplates().get(TEMPLATE_NAME).priority() == templatePriority) {
-                    logger.debug("Template [{}] already exists, skipping creation", TEMPLATE_NAME);
-                    future.complete(true);
-                    return;
+                try {
+                    String currentMapping = readIndexMappings();
```
ansjcy: We call `readIndexMappings` (which internally has IO read) multiple times and they are not all necessary. Json read happens here and inside `createTemplate`, and `createIndexAndBulk` too.

We can optimize this by potentially reading the json file in `export` function only once and passing `currentMapping` to `createTemplate` and `createIndexAndBulk`

src/test/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporterTests.java
```diff
@@ -403,4 +407,206 @@ public void testCreateIndexWithAutoExpandReplicas() throws IOException {
         // Verify there is no number_of_replicas setting.
         assertNull(settings.get("index.number_of_replicas"));
     }
+
```
ansjcy: Let's also add tests for non-cluster manager node path, there shouldn't be client interactions with index mapping creations etc.

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -328,51 +336,80 @@ String readIndexMappings() throws IOException {
...
+     * @param currentMapping The index mapping to use for template creation/comparison
      * @return CompletableFuture that completes when the template check/creation is done
      */
-    CompletableFuture<Boolean> ensureTemplateExists() {
+    CompletableFuture<Boolean> ensureTemplateExists(String currentMapping) {
         CompletableFuture<Boolean> future = new CompletableFuture<>();
+
+        // Only cluster manager node should manage templates to avoid race conditions
+        ClusterState clusterState = clusterService.state();
+        if (clusterState == null || !clusterState.nodes().isLocalNodeElectedClusterManager()) {
```
ansjcy: The integ test `testQueryInsightsExporterSettings` is failing with 
```
Index template not created within timeout
```
I did some thinking around this - this is not just a flaky test, the way we handle index template creation is not reliable. 

Currently we create the index template on demand during the export path (cluster manager only). That means if a node doesn't handle any search requests (possible if custom routing enabled, or dedicated cluster manager node), `export()` will never be triggerred on that node, thus template will never be created. This will be a big issue because data nodes will later export to indices without templates and cause data inconsistency issues.

We should consider creating index template proactively - for example, we can add a ClusterStateListener that triggers on cluster manager election and create / update index template, also trigger template check when exporter settings change (type, priority, etc.), and keep `export()` logic as-is for fallback.

Thoughts? @LilyCaroline17  @dzane17 @KishoreKicha14

LilyCaroline17: After a discussion, we have decided that instead of using an index template, we want to provide the mapping every time we create new local index.

Issue #411, opened 2025-09-05
[BUG] Missing mapping for top_queries index "source" field


dzane17:
### What is the bug?
The Query Insights local index is missing a mapping for the `source` field. OpenSearch applies dynamic mapping to fields without explicit mappings. However, because the `source` field in the local index can be very large and deeply nested, this may lead to errors such as `IllegalArgumentException` and `MapperParsingException`, as shown below:

```
java.lang.IllegalArgumentException: Limit of total fields [1000] has been exceeded
```

```
MapperParsingException[failed to parse]; nested: OpenSearchParseException[The depth of the field has exceeded the allowed limit of [20]. This limit can be set by changing the [index.mapping.depth.limit] index level setting.];
```

Sample mapping for `top_queries-2025.09.05-00381`. See `source` mapping is highly bloated.
```
{
  "top_queries-2025.09.05-00381" : {
    "mappings" : {
      "dynamic" : "true",
      "_meta" : {
        "schema_version" : 1,
        "query_insights_feature_space" : "top_n_queries"
      },
      "properties" : {
        "group_by" : {...},
        "id" : {...},
        "indices" : {...},
        "labels" : {...},
        "measurements" : {...},
        "node_id" : {...},
        "phase_latency_map" : {...},
        "search_type" : {...},
        "source" : {
          "properties" : {
            "query" : {
              "properties" : {
                "match" : {
                  "properties" : {
                    "author" : {
                      "properties" : {
                        "auto_generate_synonyms_phrase_query" : {
                          "type" : "boolean"
                        },
                        "boost" : {
                          "type" : "float"
                        },
                        "fuzzy_transpositions" : {
                          "type" : "boolean"
                        },
                        "lenient" : {
                          "type" : "boolean"
                        },
                        "max_expansions" : {
                          "type" : "long"
                        },
                        "operator" : {
                          "type" : "text",
                          "fields" : {
                            "keyword" : {
                              "type" : "keyword",
                              "ignore_above" : 256
                            }
                          }
                        },
                        "prefix_length" : {
                          "type" : "long"
                        },
                        "query" : {
                          "type" : "text",
                          "fields" : {
                            "keyword" : {
                              "type" : "keyword",
                              "ignore_above" : 256
                            }
                          }
                        },
                        "zero_terms_query" : {
                          "type" : "text",
                          "fields" : {
                            "keyword" : {
                              "type" : "keyword",
                              "ignore_above" : 256
                            }
                          }
                        }
                      }
[...]

PR #483, opened 2025-11-14
Store source field as a string in local index to optimize query storage

### LINKED ISSUE #352: [FEATURE] Optimize Storage of Large Query Sources
Currently, we store the entire query source in top queries regardless of size, potentially causing large memory footprint issues.

### What solution would you like?
- We can simply ignore storing queries with large source content, though this would result in empty query source displays in the detail page.
- Or, we can compress large query sources by extracting only essential features - specifically, store just the query shape for queries with large sources.

_A clear and concise description of any alternative solutions or features you've considered._

LilyCaroline17:
### Description
This PR addresses some bugs that we're having with how the source field is stored in the local index. The source field can be really long and/or really nested. The solution is to store the source field as a string within the local index so no exceptions related to field depth or number of fields is triggered and to maintain consistency. In-memory, it's stored as a SourceString object (with source as a string) for backwards compatibility during upgrades to detect that we are storing the source as a string instead of a Search Source Builder. 

Additionally, for backwards compatibility when existing indices already have the source field defined as an object (ie rolling upgrades), those indices will have source field written to them as SearchSourceBuilders, maintaining the existing format of source for those indices. 

### Issues Resolved
Closes #352 and #411
 along with truncation PR: #484

Dashboards PR to handle source field change: https://github.com/opensearch-project/query-insights-dashboards/pull/357

Source is in properties in the local index.
Screenshot 2025-11-14 at 12 16 15 AM

### Test Results (1000 query records)

I performed a test with 1000 query records to compare the storage usage of storing source as a string compared to storing source as a SearchSourceBuilder and got the following results: 

| Query Complexity | SearchSourceBuilder | String Storage | Memory Saved |
|------------------|-------------------|----------------|--------------|
| Small queries | 320B/record | 312B/record | **2.6%** |
| Average queries | 1.8KB/record | 751B/record | **58.4%** |
| Large queries | 4.0KB/record | 2.0KB/record | **49.3%** |

We can see that there is a significant reduction in memory storage for source when it's stored as string, especially in larger, more complex queries. 

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
@@ -14,7 +14,6 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -22,19 +21,17 @@ import org.opensearch.ExceptionsHelper;
 import org.opensearch.ResourceAlreadyExistsException;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexResponse;
-import org.opensearch.action.admin.indices.mapping.get.GetMappingsRequest;
-import org.opensearch.action.admin.indices.mapping.get.GetMappingsResponse;
 import org.opensearch.action.bulk.BulkRequestBuilder;
 import org.opensearch.action.bulk.BulkResponse;
 import org.opensearch.action.index.IndexRequest;
 import org.opensearch.cluster.ClusterState;
-import org.opensearch.cluster.metadata.MappingMetadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.xcontent.ToXContent;
+import org.opensearch.index.mapper.MapperException;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetric;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetricsCounter;
 import org.opensearch.plugin.insights.core.utils.IndexDiscoveryHelper;
@@ -128,10 +125,9 @@ public class LocalIndexExporter implements QueryInsightsExporter {
             if (!checkIndexExists(indexName)) {
                 createIndexAndBulk(indexName, records);
             } else {
-                // Index already exists, check mapping compatibility before bulk export
-                checkMappingAndBulk(indexName, records);
+                bulk(indexName, records);
             }
-        } catch (Exception e) {
+        } catch (IOException e) {
             logger.error("Unable to export query insights data:", e);
             OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.LOCAL_INDEX_EXPORTER_EXCEPTIONS);
         }
@@ -160,7 +156,7 @@ public class LocalIndexExporter implements QueryInsightsExporter {
                 if (createIndexResponse.isAcknowledged()) {
                     try {
                         bulk(indexName, records);
-                    } catch (Exception e) {
+                    } catch (IOException e) {
                         OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.LOCAL_INDEX_EXPORTER_EXCEPTIONS);
                         logger.error("Unable to index query insights data: ", e);
                     }
@@ -173,7 +169,7 @@ public class LocalIndexExporter implements QueryInsightsExporter {
                 if (cause instanceof ResourceAlreadyExistsException) {
                     try {
                         bulk(indexName, records);
-                    } catch (Exception ioe) {
+                    } catch (IOException ioe) {
                         OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.LOCAL_INDEX_EXPORTER_EXCEPTIONS);
                         logger.error("Unable to index query insights data: ", ioe);
                     }
@@ -186,57 +182,26 @@ public class LocalIndexExporter implements QueryInsightsExporter {
     }
 
     /**
-     * Check mapping compatibility and perform bulk export with appropriate format
+     * Bulk index records to the specified index with default string source serialization.
+     * This method defaults useObjectSource to false.
+     *
+     * @param indexName Name of the index to bulk records to
+     * @param records List of {@link SearchQueryRecord} to index
+     * @throws IOException If there's an error during bulk indexing
      */
-    void checkMappingAndBulk(String indexName, List<SearchQueryRecord> records) {
-        GetMappingsRequest request = new GetMappingsRequest().indices(indexName);
-
-        client.admin().indices().getMappings(request, new ActionListener<GetMappingsResponse>() {
-            @Override
-            public void onResponse(GetMappingsResponse response) {
-                boolean isObjectType = false;
-                try {
-                    MappingMetadata mappingMetadata = response.getMappings().get(indexName);
-                    if (mappingMetadata != null) {
-                        Map<String, Object> sourceMap = mappingMetadata.getSourceAsMap();
-                        Map<String, Object> properties = (Map<String, Object>) sourceMap.get("properties");
-                        if (properties != null && properties.containsKey("source")) {
-                            Map<String, Object> sourceField = (Map<String, Object>) properties.get("source");
-                            String type = (String) sourceField.get("type");
-                            isObjectType = !"text".equals(type);
-                        } else {
-                            // No explicit source mapping means dynamic mapping treats it as object
-                            isObjectType = true;
-                        }
-                    }
-                } catch (Exception e) {
-                    logger.warn("Error checking source field mapping: {}", e.getMessage());
-                }
-
-                try {
-                    bulk(indexName, records, isObjectType);
-                } catch (Exception e) {
-                    OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.LOCAL_INDEX_EXPORTER_EXCEPTIONS);
-                    logger.error("Unable to index query insights data: ", e);
-                }
-            }
-
-            @Override
-            public void onFailure(Exception e) {
-                try {
-                    bulk(indexName, records, false);
-                } catch (Exception ex) {
-                    OperationalMetricsCounter.getInstance().incrementCounter(OperationalMetric.LOCAL_INDEX_EXPORTER_EXCEPTIONS);
-                    logger.error("Unable to index query insights data: ", ex);
-                }
-            }
-        });
-    }
-
     void bulk(final String indexName, final List<SearchQueryRecord> records) throws IOException {
         bulk(indexName, records, false);
     }
 
+    /**
+     * Bulk index records to the specified index
+     *
+     * @param indexName Name of the index to bulk records to
+     * @param records List of {@link SearchQueryRecord} to index
+     * @param useObjectSource If true, serializes the SOURCE attribute as a {@link org.opensearch.search.builder.SearchSourceBuilder} object;
+     *                       if false, serializes it as a string representation {@link org.opensearch.plugin.insights.rules.model.SourceString}. Used for handling mapping conflicts.
+     * @throws IOException If there's an error during bulk indexing
+     */
     void bulk(final String indexName, final List<SearchQueryRecord> records, boolean useObjectSource) throws IOException {
         final BulkRequestBuilder bulkRequestBuilder = client.prepareBulk().setTimeout(TimeValue.timeValueMinutes(1));
         for (SearchQueryRecord record : records) {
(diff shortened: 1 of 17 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
```diff
@@ -81,7 +97,31 @@ public void consumeRecords(List<SearchQueryRecord> records) {
...
+            try {
+                source = SearchSourceBuilder.fromXContent(
+                    org.opensearch.common.xcontent.json.JsonXContent.jsonXContent.createParser(
+                        namedXContentRegistry,
+                        org.opensearch.core.xcontent.DeprecationHandler.THROW_UNSUPPORTED_OPERATION,
+                        sourceString
+                    )
+                );
+            } catch (Exception e) {
+                logger.warn("Failed to parse SOURCE string for categorization: {} - Source: {}", e.getMessage(), sourceString);
```
ansjcy: We should never log the complete source string as it will contain sensitive information

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
```diff
@@ -39,14 +39,17 @@ public final class SearchQueryCategorizer {
 
     final SearchQueryAggregationCategorizer searchQueryAggregationCategorizer;
     private static SearchQueryCategorizer instance;
+    private static org.opensearch.core.xcontent.NamedXContentRegistry namedXContentRegistry;
```
ansjcy: I see a lot of patterns like this, could you do some clean up and make sure you import them `org.opensearch.core.xcontent.NamedXContentRegistry` instead of using those ad-hoc paths?

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -138,8 +138,14 @@ public static Object readAttributeValue(StreamInput in, Attribute attribute) thr
         if (attribute == Attribute.TASK_RESOURCE_USAGES) {
             return in.readList(TaskResourceInfo::readFromStream);
         } else if (attribute == Attribute.SOURCE) {
-            SearchSourceBuilder builder = new SearchSourceBuilder(in);
-            return builder;
+            Object value = in.readGenericValue();
```
ansjcy: This will potentially cause a critical node failure during rolling upgrade, we should be extra careful when modifying the serialization logic.

When rolling upgrade happens, say node A is on 3.3 or lower and node B is on 3.4+ and has this change.
During upgrade, when node A is the coordinator node, node B will create a SearchQueryRecord with string as source, and send back to node A. When node A deserializes, it will call `SearchSourceBuilder builder = new SearchSourceBuilder(in);`, but this expect `SearchSourceBuilder` binary format thus the node will crash!

You should do version checks here, similar to https://github.com/opensearch-project/query-insights/blob/34391ac34311dfeadc4d7e526c561263d80bcdba/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java#L149

When sending and receiving data in `readAttributeValue` and `writeValueTo`, we should check the out.getVersion and parse to `SearchSourceBuilder` if <= 3.3

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -217,10 +220,26 @@ public void onFailure(Exception e) {
     void bulk(final String indexName, final List<SearchQueryRecord> records) throws IOException {
         final BulkRequestBuilder bulkRequestBuilder = client.prepareBulk().setTimeout(TimeValue.timeValueMinutes(1));
         for (SearchQueryRecord record : records) {
-            bulkRequestBuilder.add(
-                new IndexRequest(indexName).id(record.getId())
-                    .source(record.toXContentForExport(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS))
-            );
+            Object sourceValue = record.getAttributes().get(Attribute.SOURCE);
+            if (sourceValue != null && !(sourceValue instanceof String)) {
```
ansjcy: Could you explain on when will this`if (sourceValue != null && !(sourceValue instanceof String))` be true? This if statement looks like a dead code to me.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -259,8 +262,22 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
...
+                                    DeprecationHandler.THROW_UNSUPPORTED_OPERATION,
+                                    sourceString
+                                );
+                                attributes.put(Attribute.SOURCE, SearchSourceBuilder.fromXContent(sourceParser, false));
+                            } catch (Exception e) {
+                                attributes.put(Attribute.SOURCE, sourceString);
+                            }
+                        } else {
+                            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
+                            attributes.put(Attribute.SOURCE, SearchSourceBuilder.fromXContent(parser, false));
```
dzane17: I thought this PR is converting the in-memory representation of source to `String`. What scenario do we need `SearchSourceBuilder.fromXContent`?

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -138,8 +145,32 @@ public static Object readAttributeValue(StreamInput in, Attribute attribute) thr
...
+            if (in.getVersion().onOrAfter(Version.V_3_4_0)) {
+                // For 3.4+ versions, always return as string
+                Object value = in.readGenericValue();
+                return value instanceof String ? value : value.toString();
+            } else {
+                // For pre-3.4 versions, we need to handle both old format (SearchSourceBuilder)
+                // and new format (string from 3.4+ nodes) gracefully
+
+                // Mark the current position in case we need to reset
+                in.mark(1024); // Mark with a reasonable read-ahead limit
```
ansjcy: I don't think all StreamInput supports mark and reset, and will potentially cause exception. any reason we cannot read it only once without relying on mark and reset? Also the 1024 byte limit is arbitrary - A `SearchSourceBuilder` can easily exceed this limit, causing the reset() to fail.

ansjcy:
Also tests are still failing, could you make sure the pipeline is green?

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCategorizer.java
```diff
@@ -81,7 +98,25 @@ public void consumeRecords(List<SearchQueryRecord> records) {
      * @param record search query record
      */
     public void categorize(SearchQueryRecord record) {
-        SearchSourceBuilder source = (SearchSourceBuilder) record.getAttributes().get(Attribute.SOURCE);
+        SearchSourceBuilder source = record.getSearchSourceBuilder();
+        if (source == null) {
+            // SearchSourceBuilder should always be available but as a fallback try to parse from string source
+            String sourceStr = (String) record.getAttributes().get(Attribute.SOURCE);
```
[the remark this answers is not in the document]
LilyCaroline17: I wanted this as a failsafe just in case, but I can remove it.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
LilyCaroline17: That makes sense, for older versions I don't think we would be dealing with strings at all so I'll just remove it completely and read as a SearchSourceBuilder. However, I did realize that for nodes with versions 3.4+, they may read an index with old data that has source stored as a SearchSourceBuilder, so that scenario needs to be accounted for because doing in.readGenericValue on a SearchSourceBuilder would probably result in an exception.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -118,7 +119,13 @@ public static void writeValueTo(StreamOutput out, Object attributeValue) throws
         if (attributeValue instanceof List) {
             out.writeList((List<? extends Writeable>) attributeValue);
         } else if (attributeValue instanceof SearchSourceBuilder) {
-            ((SearchSourceBuilder) attributeValue).writeTo(out);
+            if (out.getVersion().onOrAfter(Version.V_3_4_0)) {
```
dzane17: Don't forget to change all version checks to 3.5 since 3.4 deadline has passed https://opensearch.org/releases/

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -118,7 +119,13 @@ public static void writeValueTo(StreamOutput out, Object attributeValue) throws
         if (attributeValue instanceof List) {
             out.writeList((List<? extends Writeable>) attributeValue);
         } else if (attributeValue instanceof SearchSourceBuilder) {
```
dzane17: The if condition on this line should never be true. Do we need a custom writer for the source attribute? There is a slight difference in how source is currently being handled vs the generic "else" case.
```
out.writeGenericValue(attributeValue.toString());

// else case on line 132
out.writeGenericValue(attributeValue);
```

LilyCaroline17: Yes, Chenyang also mentioned this, I'm removing this if else within the else if statement.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -508,12 +553,38 @@ public XContentBuilder toXContent(final XContentBuilder builder, final Params pa
...
+        throws IOException {
         builder.startObject();
         builder.field("timestamp", timestamp);
         builder.field("id", id);
 
         for (Map.Entry<Attribute, Object> entry : attributes.entrySet()) {
-            builder.field(entry.getKey().toString(), entry.getValue());
+            if (entry.getKey() == Attribute.SOURCE && useObjectSource) {
+                if (searchSourceBuilder != null) {
+                    builder.field(entry.getKey().toString(), searchSourceBuilder);
```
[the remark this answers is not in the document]
LilyCaroline17: This is for rolling upgrades, so useObjectSource is a boolean that tracks if the index we're writing to doesn't have the source mapping yet so it expects source as a SearchSourceBuilder. If so, we have to write source as a SearchSourceBuilder in order to successfully write to that index.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
dzane17: @ansjcy @LilyCaroline17 Actually this method is very complicated. For out.version < 3.4 we need to convert the source String to SearchSourceBuilder. But there is no way to detect source field is being process inside this method.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
LilyCaroline17: Yes that makes sense! writeValueTo does a generic write for source as a string. My only concern is that with rolling upgrades, an updated node may write to an existing index, so then it has to write source as a SearchSourceBuilder in order for it to work, meaning that it would reach the else case, which would break as we wouldn't be able to read the SearchSourceBuilder with the in.readGenericValue().

src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
```diff
@@ -180,12 +185,64 @@ public void onFailure(Exception e) {
         });
     }
 
+    /**
+     * Check mapping compatibility and perform bulk export with appropriate format
+     */
+    void checkMappingAndBulk(String indexName, List<SearchQueryRecord> records) {
```
dzane17: Right now we are checking the index mapping before every bulk, but most times this is unnecessary work. I wonder if it would make more sense to first try the bulk request with useObjectSource = true. Only if the bulk fails with `MapperException` then we optionally do the "check mapping" flow and retry with useObjectSource = false.

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -508,12 +551,39 @@ public XContentBuilder toXContent(final XContentBuilder builder, final Params pa
      * @throws IOException if an I/O error occurs during serialization.
      */
     public XContentBuilder toXContentForExport(final XContentBuilder builder, final Params params) throws IOException {
+        return toXContentForExport(builder, params, false);
+    }
+
+    public XContentBuilder toXContentForExport(final XContentBuilder builder, final Params params, boolean useObjectSource)
```
dzane17: nit: add javadoc to explain useObjectSource

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -508,12 +551,39 @@ public XContentBuilder toXContent(final XContentBuilder builder, final Params pa
...
         builder.startObject();
         builder.field("timestamp", timestamp);
         builder.field("id", id);
 
         for (Map.Entry<Attribute, Object> entry : attributes.entrySet()) {
-            builder.field(entry.getKey().toString(), entry.getValue());
+            if (entry.getKey() == Attribute.SOURCE && useObjectSource) {
+                if (searchSourceBuilder != null) {
+                    builder.field(entry.getKey().toString(), searchSourceBuilder);
+                } else {
```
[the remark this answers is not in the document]
LilyCaroline17: That makes sense. I had it as a fallback but searchSourceBuilder is always non-null for export as that means it hasn't reached local index yet.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -127,6 +132,28 @@ static void writeTo(final StreamOutput out, final Attribute attribute) throws IO
...
+                        parser.close();
+                    } else {
+                        new SearchSourceBuilder().writeTo(out);
+                    }
+                } catch (Exception e) {
+                    // Unable to convert source to SearchSourceBuilder, sending dummy object instead
+                    new SearchSourceBuilder().writeTo(out);
+                }
+            }
         } else if (attributeValue instanceof SearchSourceBuilder) {
```
dzane17: This case `instanceof SearchSourceBuilder` can be deleted

ansjcy:
This PR to fix the flaky testTopQueriesResponses is merged, you can try rebasing now https://github.com/opensearch-project/query-insights/pull/513

ansjcy (approved):

PR #484, opened 2025-11-17
Truncate source string in local index to optimize query storage

### LINKED ISSUE #352: [FEATURE] Optimize Storage of Large Query Sources

LilyCaroline17:
### Description
This PR addresses some bugs that we're having with how the source field is stored in the local index. The source field can be really long even after being converted into string format. The solution is to truncate the source string to reduce the storage size needed for queries. 

### Issues Resolved
Closes #352 and #411 in tandem with #483. 

#483 is required prior to merging this PR. 

Truncation (set to 100 characters)
Screenshot 2025-11-17 at 3 06 53 PM

### DIFF
--- a/src/test/java/org/opensearch/plugin/insights/QueryInsightsTestUtils.java
+++ b/src/test/java/org/opensearch/plugin/insights/QueryInsightsTestUtils.java
@@ -166,6 +166,7 @@ final public class QueryInsightsTestUtils {
             Map<Attribute, Object> attributes = new HashMap<>();
             attributes.put(Attribute.SEARCH_TYPE, SearchType.QUERY_THEN_FETCH.toString().toLowerCase(Locale.ROOT));
             attributes.put(Attribute.SOURCE, new SourceString(searchSourceBuilder.toString()));
+            attributes.put(Attribute.SOURCE_TRUNCATED, false);
             attributes.put(Attribute.TOTAL_SHARDS, randomIntBetween(1, 100));
             attributes.put(Attribute.INDICES, randomArray(1, 3, Object[]::new, () -> randomAlphaOfLengthBetween(5, 10)));
             attributes.put(Attribute.PHASE_LATENCY_MAP, phaseLatencyMap);


src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
```diff
@@ -154,6 +154,7 @@ public List<Setting<?>> getSettings() {
             QueryInsightsSettings.TOP_N_EXPORTER_TYPE,
             QueryInsightsSettings.TOP_N_EXPORTER_TEMPLATE_PRIORITY,
             QueryInsightsSettings.TOP_N_QUERIES_EXCLUDED_INDICES,
+            QueryInsightsSettings.MAX_SOURCE_LENGTH,
```
ansjcy: Let's make the naming consistent, `TOP_N_QUERIES_ MAX_SOURCE_LENGTH `

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -259,9 +266,31 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
...
+                                    attributes.put(Attribute.SOURCE, sourceString);
+                                }
+                            }
+                        } else {
+                            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
+                            attributes.put(Attribute.SOURCE, SearchSourceBuilder.fromXContent(parser, false));
+                        }
                         break;
+                    case SOURCE_TRUNCATED:
+                        attributes.put(Attribute.SOURCE_TRUNCATED, parser.booleanValue());
```
ansjcy: you might be missing a break here

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
```diff
@@ -326,6 +326,22 @@ public class QueryInsightsSettings {
...
+     */
+    public static final int DEFAULT_MAX_SOURCE_LENGTH = 10000;
+
+    /**
+     * Setting for maximum source length before truncation
+     */
+    public static final Setting<Integer> MAX_SOURCE_LENGTH = Setting.intSetting(
+        TOP_N_QUERIES_SETTING_PREFIX + ".max_source_length",
+        DEFAULT_MAX_SOURCE_LENGTH,
+        1, // min value is 1
```
ansjcy: Also, 1 byte as the min value does not make sense to me, even the simplest query `"{}"` has 2 bytes

src/test/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListenerTests.java
```diff
@@ -463,4 +463,72 @@ public void testExcludedIndicesValidation() {
...
+        QueryInsightsListener queryInsightsListener = new QueryInsightsListener(clusterService, queryInsightsService);
+        queryInsightsListener.setMaxSourceLength(100);
+
+        SearchSourceBuilder longSource = new SearchSourceBuilder();
+        StringBuilder longQuery = new StringBuilder("{\"query\":{\"bool\":{\"must\":[");
+        for (int i = 0; i < 50; i++) {
+            if (i > 0) longQuery.append(",");
+            longQuery.append("{\"match\":{\"field").append(i).append("\":\"value").append(i).append("\"}}");
+        }
+        longQuery.append("]}}");
```
ansjcy: Is this longQuery used anywhere??

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -313,7 +327,20 @@ private void constructSearchQueryRecord(final SearchPhaseContext context, final
 
             Map<Attribute, Object> attributes = new HashMap<>();
             attributes.put(Attribute.SEARCH_TYPE, request.searchType().toString().toLowerCase(Locale.ROOT));
-            attributes.put(Attribute.SOURCE, request.source());
+            // Handle source with truncation for very long queries
+            if (request.source() != null) {
+                String sourceString = request.source().toString();
+                if (sourceString.length() > maxSourceLength) {
+                    attributes.put(Attribute.SOURCE, sourceString.substring(0, maxSourceLength));
```
ansjcy: Can we use a more sophisticated logic for truncating rather than simply do `substring(0, maxSourceLength)` ? substring will make the truncated string guranteed not able to parse back.

I was thinking about extracting the shape until it goes below the threshold, but if it is too complicated, we can do something like below at least to add the closing braces and `"` at least (written by Claude so do double check)

```java
private String truncateJsonSafely(String json, int maxLength) {
    if (json.length() <= maxLength) return json;

    String truncated = json.substring(0, maxLength);

    // Count unmatched braces/brackets
    int braces = 0, brackets = 0;
    boolean inString = false;
    boolean escaped = false;

    for (int i = 0; i < truncated.length(); i++) {
        char c = truncated.charAt(i);
        if (escaped) {
            escaped = false;
            continue;
        }
        if (c == '\\') {
            escaped = true;
        } else if (c == '"') {
            inString = !inString;
        } else if (!inString) {
            if (c == '{') braces++;
            else if (c == '}') braces--;
            else if (c == '[') brackets++;
            else if (c == ']') brackets--;
        }
    }

    // Close unclosed structures
    StringBuilder result = new StringBuilder(truncated);
    while (brackets > 0) { result.append(']'); brackets--; }
    while (braces > 0) { result.append('}'); braces--; }

    return result.toString();
```

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -259,9 +266,31 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
...
+                            if (Boolean.TRUE.equals(isTruncated)) {
+                                attributes.put(Attribute.SOURCE, sourceString);
+                            } else {
+                                try {
+                                    XContentParser sourceParser = JsonXContent.jsonXContent.createParser(
+                                        NamedXContentRegistry.EMPTY,
+                                        DeprecationHandler.THROW_UNSUPPORTED_OPERATION,
+                                        sourceString
+                                    );
+                                    attributes.put(Attribute.SOURCE, SearchSourceBuilder.fromXContent(sourceParser, false));
```
ansjcy: I thought we want to store `SOURCE` field as string, why do you need to store it as `SearchSourceBuilder` here? 

considering my comment here https://github.com/opensearch-project/query-insights/pull/483/files#r2536620414, and the challenge to read `SOURCE_TRUNCATED` potentially before it is parsed, you can always do `attributes.put(Attribute.SOURCE, sourceString);` here, and after the `switch`, based on the `Attribute.SOURCE_TRUNCATED` value, parse the string to `SearchSourceBuilder` and store into the new SearchQueryRecord field if possible

src/main/java/org/opensearch/plugin/insights/settings/QueryInsightsSettings.java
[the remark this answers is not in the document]
dzane17: Max should be around 100,000

dzane17: I believe the setting is on character count not bytes. "length" implies characters as opposed to "size." but wondering if we should change the setting and variable names to clarify? At least let's update the comments.

src/main/java/org/opensearch/plugin/insights/core/service/grouper/MinMaxHeapQueryGrouper.java
```diff
@@ -15,6 +15,8 @@
 import org.apache.logging.log4j.LogManager;
```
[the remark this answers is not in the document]
LilyCaroline17: the unit tests are in MinMaxHeapQueryGrouperTests.java

PR #494, opened 2025-12-02
[Feat] Add Live Queries Cache with Task Resource Aggregation and Configurable Idle Timeout

KishoreKicha14:
### Description

This PR implements a live queries cache feature that polls running search queries with task resource aggregation and provides real-time query monitoring capabilities. The cache includes configurable idle timeout with automatic stop functionality to optimize resource usage.

#### Live Queries Cache Implementation

- **New Class**: `LiveQueriesCache` - Polls running search tasks every 1 second
- Maintains top 1000 queries sorted by latency, returns top 100
- Extracts comprehensive query metadata:
  - Latency, CPU, and memory measurements with task resource aggregation
  - Task resource usage information from all search phases
  - Workload group ID (WLM integration)
  - User information (username and roles via `PrincipalExtractor`)
  - Query description and cancellation status
  - Node ID

#### Task Resource Aggregation
- Aggregates CPU and memory usage from task resource stats
- Extracts `total` resource usage from `TaskResourceStats`
- Provides accurate resource consumption metrics for running queries
- Includes detailed `TaskResourceInfo` in response with action, task IDs, and resource usage

#### New API Parameter: `use_live_cache`

The `/_insights/live_queries` API now supports a `use_live_cache` parameter:

**When `use_live_cache=true`**: Returns queries from the in-memory cache (updated every 1 second)
- Faster response times - no real-time task list API call
- Suitable for monitoring dashboards that poll frequently
- Cache starts automatically on first use

**When `use_live_cache=false` or omitted**: Makes real-time `ListTasksRequest` call
- Returns current running queries directly from task manager
- Default behavior (backward compatible)

#### Configurable Idle Timeout Setting
- **Setting**: `search.insights.live_queries.cache.idle_timeout`
- **Default**: 5 minutes
- **Range**: 2-10 minutes
- **Special Value**: `-1` to disable cache (throws error when attempting to use)
- **Type**: Dynamic cluster setting

#### Cache Behavior
- Cache starts lazily on first `use_live_cache=true` API call
- Polls running tasks every 1 second using `ListTasksRequest` with detailed stats
- Automatically stops after configured idle period with no API requests
- When disabled (`-1`), API returns clear error message
- Cache clears memory on stop to free resources

#### Example 1: Get live queries using cache (fast, cached response)
```bash
curl -X GET "http://localhost:9200/_insights/live_queries?use_live_cache=true&pretty"
{
  "live_queries" : [
    {
      "timestamp" : 1769207048182,
      "id" : "KBK9GE-vRO2gJHiFyVAFjw:64478",
      "node_id" : "KBK9GE-vRO2gJHiFyVAFjw",
      "description" : "indices[test-index], search_type[QUERY_THEN_FETCH], source[{\"size\":1000,\"query\":{\"match_all\":{\"boost\":1.0}}}]",
      "wlm_group_id" : null,
      "task_resource_usages" : [
        {
          "taskId" : 64478,
          "parentTaskId" : -1,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 0,
            "memory_in_bytes" : 0
          }
        },
        {
          "action" : "indices:data/read/search[phase/query]",
          "taskId" : 64479,
          "parentTaskId" : 64478,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 0,
            "memory_in_bytes" : 0
          }
        }
      ],
      "is_cancelled" : false,
      "measurements" : {
        "memory" : {
          "number" : 0,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 0,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "latency" : {
          "number" : 528502625,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    },
    {
      "timestamp" : 1769207048187,
      "id" : "KBK9GE-vRO2gJHiFyVAFjw:64480",
      "node_id" : "KBK9GE-vRO2gJHiFyVAFjw",
      "description" : "indices[test-index], search_type[QUERY_THEN_FETCH], source[{\"size\":1000,\"query\":{\"match_all\":{\"boost\":1.0}}}]",
      "wlm_group_id" : null,
      "task_resource_usages" : [
        {
          "taskId" : 64480,
          "parentTaskId" : -1,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 0,
            "memory_in_bytes" : 0
          }
        },
        {
          "action" : "indices:data/read/search[phase/query]",
          "taskId" : 64481,
          "parentTaskId" : 64480,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 0,
            "memory_in_bytes" : 0
          }
        }
      ],
      "is_cancelled" : false,
      "measurements" : {
        "memory" : {
          "number" : 0,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 0,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "latency" : {
          "number" : 522958042,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}
```
Example 2: Get live queries without cache (real-time, direct from task manager)

```
curl -X GET "http://localhost:9200/_insights/live_queries?pretty"

```
Example 3: Set idle timeout to 2 minutes
```curl -X PUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent": {
    "search.insights.live_queries.cache.idle_timeout": "2m"
  }
}'
```
```
Response:

{
  "acknowledged": true,
  "persistent": {
    "search": {
      "insights": {
        "live_queries": {
          "cache": {
            "idle_timeout": "2m"
          }
        }
      }
    }
  },
  "transient": {}
}
```
Example 4: Cache auto-stops after idle timeout

 First request - cache starts
```curl -X GET "http://localhost:9200/_insights/live_queries?use_live_cache=true&pretty"```
 Returns queries...

Wait 2+ minutes (if timeout set to 2m)

Second request - cache has stopped, returns empty
```curl -X GET "http://localhost:9200/_insights/live_queries?use_live_cache=true&pretty"```

Response after idle timeout:
```
{
  "live_queries": []
}
```

Example 5: Disable cache
```curl -X PUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent": {
    "search.insights.live_queries.cache.idle_timeout": "-1"
  }
}'
```

Response:
```
{
  "acknowledged": true,
  "persistent": {
    "search": {
      "insights": {
        "live_queries": {
          "cache": {
            "idle_timeout": "-1"
          }
        }
      }
    }
  },
  "transient": {}
}
```

Example 6: Error when cache is disabled

```
curl -X GET "http://localhost:9200/_insights/live_queries?use_live_cache=true&pretty"
```
Response:
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_state_exception",
        "reason": "Live queries cache is disabled. Set search.insights.live_queries.cache.idle_timeout to a positive value (2-10 minutes) to enable it."
      }
    ],
    "type": "illegal_state_exception",
    "reason": "Live queries cache is disabled. Set search.insights.live_queries.cache.idle_timeout to a positive value (2-10 minutes) to enable it."
  },
  "status": 500
}
```
without use_live_cache
```
 curl -s "http://localhost:9200/_insights/live_queries?pretty" | head -80
{
  "live_queries" : [
    {
      "timestamp" : 1769213176483,
      "id" : "KBK9GE-vRO2gJHiFyVAFjw:11130",
      "task_resource_usages" : [
        {
          "taskId" : 11130,
          "parentTaskId" : -1,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 0,
            "memory_in_bytes" : 0
          }
        },
        {
          "action" : "indices:data/read/search[phase/query]",
          "taskId" : 11131,
          "parentTaskId" : 11130,
          "nodeId" : "KBK9GE-vRO2gJHiFyVAFjw",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 7050000,
            "memory_in_bytes" : 3733696
          }
        }
      ],
      "is_cancelled" : false,
      "description" : "indices[test-index], search_type[QUERY_THEN_FETCH], source[{\"size\":1000,\"query\":{\"match_all\":{\"boost\":1.0}}}]",
      "wlm_group_id" : "DEFAULT_WORKLOAD_GROUP",
      "node_id" : "KBK9GE-vRO2gJHiFyVAFjw",
      "measurements" : {
        "memory" : {
          "number" : 3733696,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "cpu" : {
          "number" : 7050000,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "latency" : {
          "number" : 2014466583,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
}
```

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -74,6 +74,20 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
     private final PrincipalExtractor principalExtractor;
 
     /**
+     * Extract and store user information for the given search task
+     */
+    private void extractAndStoreUserInfo(SearchTask searchTask) {
+        if (principalExtractor != null) {
+            PrincipalExtractor.UserPrincipalInfo userInfo = principalExtractor.extractUserInfo();
+            if (userInfo != null && userInfo.getUserName() != null) {
+                String taskId = clusterService.localNode().getId() + ":" + searchTask.getId();
+                String[] roles = userInfo.getRoles() != null ? userInfo.getRoles().toArray(new String[0]) : null;
+                queryInsightsService.getLiveQueriesCache().storeTaskUserInfo(taskId, userInfo.getUserName(), roles);
+            }
+        }
+    }
+
+    /**
      * Constructor for QueryInsightsListener
      *
      * @param clusterService       The Node's cluster service.
@@ -243,7 +257,11 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
     }
 
     @Override
-    public void onPhaseStart(SearchPhaseContext context) {}
+    public void onPhaseStart(SearchPhaseContext context) {
+        if (context.getTask() instanceof SearchTask) {
+            extractAndStoreUserInfo((SearchTask) context.getTask());
+        }
+    }
 
     @Override
     public void onPhaseEnd(SearchPhaseContext context, SearchRequestContext searchRequestContext) {}
--- a/src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
@@ -1,0 +1,223 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.plugin.insights.core.service;
+
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.LIVE_QUERIES_MAX_CACHE_SIZE;
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.LIVE_QUERIES_MAX_RETURNED_QUERIES;
+import static org.opensearch.plugin.insights.settings.QueryInsightsSettings.LIVE_QUERIES_POLLING_INTERVAL_MS;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.admin.cluster.node.tasks.list.ListTasksRequest;
+import org.opensearch.action.admin.cluster.node.tasks.list.ListTasksResponse;
+import org.opensearch.cluster.node.DiscoveryNode;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.plugin.insights.rules.model.MetricType;
+import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
+import org.opensearch.threadpool.Scheduler;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.TransportService;
+import org.opensearch.transport.client.Client;
+
+/**
+ * Cache for live running queries
+ */
+public class LiveQueriesCache {
+
+    private static final Logger logger = LogManager.getLogger(LiveQueriesCache.class);
+    private static final String SEARCH_ACTION = "indices:data/read/search*";
+    private static final AtomicReference<LiveQueriesCache> INSTANCE = new AtomicReference<>();
+
+    private final AtomicReference<SearchQueryRecord[]> sortedQueries = new AtomicReference<>(new SearchQueryRecord[0]);
+    private final Client client;
+    private final ThreadPool threadPool;
+    private final TransportService transportService;
+    private volatile Scheduler.Cancellable pollingTask;
+    private final Map<String, Map<String, Object>> taskUserInfoCache = new ConcurrentHashMap<>();
+    private final AtomicBoolean started = new AtomicBoolean(false);
+    private final AtomicLong lastAccessTime = new AtomicLong(0);
+    private final AtomicReference<Scheduler.Cancellable> idleCheckTask = new AtomicReference<>();
+    private final long idleTimeoutMs;
+    private static final TimeValue IDLE_CHECK_INTERVAL = new TimeValue(60, TimeUnit.SECONDS);
+
+    public static LiveQueriesCache getInstance(
+        Client client,
+        ThreadPool threadPool,
+        TransportService transportService,
+        long idleTimeoutMs
+    ) {
+        if (idleTimeoutMs < 0) {
+            throw new IllegalStateException(
+                "Live queries cache is disabled. Set search.insights.live_queries.cache_timeout to a positive value (2-10 minutes) to enable it."
+            );
+        }
+        LiveQueriesCache cache = INSTANCE.get();
+        if (cache == null) {
+            LiveQueriesCache newCache = new LiveQueriesCache(client, threadPool, transportService, idleTimeoutMs);
+            if (!INSTANCE.compareAndSet(null, newCache)) {
+                newCache = INSTANCE.get();
+            }
+            cache = newCache;
+        }
+        cache.recordAccess();
+        cache.start();
+        return cache;
+    }
+
+    public static void stopInstance() {
+        LiveQueriesCache cache = INSTANCE.get();
+        if (cache != null) {
+            cache.stop();
+        }
+    }
+
+    public LiveQueriesCache(Client client, ThreadPool threadPool, TransportService transportService, long idleTimeoutMs) {
+        this.client = client;
+        this.threadPool = threadPool;
+        this.transportService = transportService;
+        this.idleTimeoutMs = idleTimeoutMs;
+    }
+
+    public void start() {
+        if (!started.compareAndSet(false, true)) {
+            return;
+        }
+        lastAccessTime.set(System.currentTimeMillis());
+        pollingTask = threadPool.scheduleWithFixedDelay(
+            this::getRunningTasks,
+            new TimeValue(LIVE_QUERIES_POLLING_INTERVAL_MS, TimeUnit.MILLISECONDS),
+            ThreadPool.Names.GENERIC
+        );
+        getRunningTasks();
+        startIdleCheck();
+    }
+
+    public void stop() {
+        if (!started.compareAndSet(true, false)) {
+            return;
+        }
+        if (pollingTask != null) {
+            pollingTask.cancel();
+            pollingTask = null;
+        }
+        Scheduler.Cancellable task = idleCheckTask.getAndSet(null);
+        if (task != null) {
+            task.cancel();
+        }
+        lastAccessTime.set(0);
+    }
+
+    public boolean isStarted() {
+        return started.get();
+    }
+
+    public void recordAccess() {
+        lastAccessTime.set(System.currentTimeMillis());
+    }
+
+    private void startIdleCheck() {
+        if (idleTimeoutMs < 0) {
+            return;
+        }
+        Scheduler.Cancellable task = threadPool.scheduleWithFixedDelay(() -> {
+            if (started.get() && System.currentTimeMillis() - lastAccessTime.get() > idleTimeoutMs) {
+                stop();
+            }
+        }, IDLE_CHECK_INTERVAL, ThreadPool.Names.GENERIC);
+        idleCheckTask.set(task);
+    }
+
+    private void getRunningTasks() {
+        try {
+            ListTasksRequest request = new ListTasksRequest().setActions(SEARCH_ACTION).setDetailed(true);
+
+            // Call listTasks directly on local node instead of using cluster().listTasks() to avoid fan-out
+            DiscoveryNode localNode = transportService.getLocalNode();
+            if (localNode != null) {
+                request.setNodes(new String[] { localNode.getId() });
+
+                client.admin().cluster().listTasks(request, new ActionListener<ListTasksResponse>() {
+                    @Override
+                    public void onResponse(ListTasksResponse response) {
+                        try {
+                            Map<String, SearchQueryRecord> parentRecords = LiveQueriesHelper.processTaskGroups(
+                                response.getTaskGroups(),
+                                transportService,
+                                true,
+                                taskUserInfoCache,
+                                null,
+                                null
+                            );
+
+                            // Clean up user info cache for tasks that are no longer running
+                            Set<String> activeTaskIds = parentRecords.keySet();
+                            taskUserInfoCache.keySet().removeIf(taskId -> !activeTaskIds.contains(taskId));
+
+                            TreeSet<SearchQueryRecord> allQueries = new TreeSet<>((a, b) -> {
+                                int latencyCompare = Long.compare(
+                                    ((Number) b.getMeasurement(MetricType.LATENCY)).longValue(),
+                                    ((Number) a.getMeasurement(MetricType.LATENCY)).longValue()
+                                );
+                                return latencyCompare != 0 ? latencyCompare : a.getId().compareTo(b.getId());
+                            });
+
+                            for (SearchQueryRecord record : parentRecords.values()) {
+                                allQueries.add(record);
+                                if (allQueries.size() > LIVE_QUERIES_MAX_CACHE_SIZE) {
+                                    allQueries.pollLast();
+                                }
+                            }
+
+                            sortedQueries.set(
+                                allQueries.stream().limit(LIVE_QUERIES_MAX_RETURNED_QUERIES).toArray(SearchQueryRecord[]::new)
+                            );
+                        } catch (Throwable e) {
+                            logger.error("Error processing live queries response: {}", e.getMessage());
+                        }
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        logger.debug("Failed to list tasks during polling: {}", e.getMessage());
+                    }
+                });
+            }
+        } catch (Throwable e) {
+            logger.debug("Error polling running tasks: {}", e.getMessage());
+        }
+    }
+
+    public List<SearchQueryRecord> getCurrentQueries() {
+        return Arrays.asList(sortedQueries.get());
+    }
+
+    public void storeTaskUserInfo(String taskId, String username, String[] roles) {
+        Map<String, Object> userInfo = new HashMap<>();
+        userInfo.put("username", username);
+        userInfo.put("roles", roles);
+        taskUserInfoCache.put(taskId, userInfo);
+    }
+
+    public void removeTaskUserInfo(String taskId) {
+        taskUserInfoCache.remove(taskId);
+    }
+
+}
--- a/src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesRequest.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesRequest.java
@@ -29,6 +29,8 @@ public class LiveQueriesRequest extends ActionRequest {
     // Node IDs to filter queries by
     private final String[] nodeIds;
     private String wlmGroupId;
+    private String taskId;
+    private final boolean useLiveCache;
 
     /**
      * Constructor for LiveQueriesRequest
@@ -45,6 +47,13 @@ public class LiveQueriesRequest extends ActionRequest {
         if (in.getVersion().onOrAfter(Version.V_3_3_0)) {
             this.wlmGroupId = in.readOptionalString();
         }
+        if (in.getVersion().onOrAfter(Version.V_3_5_0)) {
+            this.taskId = in.readOptionalString();
+            this.useLiveCache = in.readBoolean();
+            in.readBoolean(); // Read and discard includeFinished for backwards compatibility
+        } else {
+            this.useLiveCache = false;
+        }
     }
 
     /**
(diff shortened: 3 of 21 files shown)


dzane17 (commented):
Looks like you are adding a new stream API. Why not replace the existing API with this cache implementation?

src/main/java/org/opensearch/plugin/insights/rules/resthandler/live_queries/RestLiveQueriesStreamAction.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Rest action for streaming live queries
+ */
+public class RestLiveQueriesStreamAction extends BaseRestHandler {
+
+    @Override
+    public List<Route> routes() {
+        return List.of(
+            new Route(GET, LIVE_QUERIES_BASE_URI + "/stream"),
+            new Route(GET, String.format(Locale.ROOT, "%s/stream/{nodeId}", LIVE_QUERIES_BASE_URI))
```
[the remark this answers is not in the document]
KishoreKicha14: updated

KishoreKicha14:
The cache uses a bounded priority queue (min-heap) to efficiently maintain the top 100 longest-running queries. This approach provides O(100) memory usage instead of storing all running queries - for example, if 10,000 queries are running, only 100 are kept in memory. Each insertion takes O(log 100) constant time, and no full sort is needed since the min-heap keeps the shortest query at the top and evicts it when the size exceeds 100.

Limitation with this approach when handling user-specfic filtering:

If a user's specific query isn't in the top 100 cluster-wide, it won't appear in cached mode.
Example: User C's query on node3 with wlm-batch has 50ms latency, but all top 100 queries have >100ms latency. User C won't see their query in cached mode even with node_id=node3 filter.

can we increase cache size to 1000 or 5000 and add hybrid fallback to direct mode when filters return empty.

@dzane17 any thoughts on this

KishoreKicha14:
Screenshot 2025-12-09 at 3 20 49 AM

KishoreKicha14:
## **Live Queries — Default Values**

| **Configuration**        | **Value**            |
|--------------------------|----------------------|
| Max Cache Size           | 1000                 |
| Max Returned Queries     | 100                  |
| Initial Startup Delay    | 5 seconds            |
| Polling Interval         | 100 milliseconds     |
| Inactivity Timeout       | 5 minutes        |

## **Finished Queries — Default Values**

| **Configuration**        | **Value**        |
|--------------------------|------------------|
| Max Cache Size           | 1000             |
| Max Returned Queries     | 50               |
| Retention Period         | 30 seconds       |
| Inactivity Timeout       | 5 minutes        |

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,199 @@
...
+    public void stop() {
+        if (pollingTask != null) {
+            pollingTask.cancel();
+        }
+    }
+
+    private void pollRunningTasks() {
+        try {
+            ListTasksRequest request = new ListTasksRequest();
+            request.setActions(SEARCH_ACTION);
```
dzane17: Combine these lines `ListTasksRequest().setActions(SEARCH_ACTION)`

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,199 @@
...
+
+            client.admin().cluster().listTasks(request, new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
+                    try {
+                        logger.info("LiveQueriesCache polling found {} total tasks", response.getTasks().size());
+                        Map<String, TaskInfo> currentTasks = new java.util.HashMap<>();
```
dzane17: `new HashMap<>()`. Same on next line

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,199 @@
...
+        // Add initial delay to allow cluster initialization, then poll every 10ms
+        pollingTask = threadPool.schedule(this::startPolling, new TimeValue(5, TimeUnit.SECONDS), ThreadPool.Names.GENERIC);
+    }
+
+    private void startPolling() {
+        try {
+            // Check if cluster is ready before starting polling
+            client.admin().cluster().listTasks(new ListTasksRequest(), new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
```
dzane17: We are ignoring this ListTasksResponse completely. 

Is it even necessary to call `listTasks` inside `startPolling()`? I think we should schedule the recurring job and separately call `pollRunningTasks()` once here or in a parent method.

src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesRequest.java
```diff
@@ -111,6 +114,30 @@ public String getWlmGroupId() {
         return wlmGroupId;
     }
 
+    public boolean isCached() {
```
ansjcy: do we need to serialize them in the writeTo function?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -223,10 +225,12 @@ private void updateQueryInsightsState() {
 
         if (anyFeatureEnabled && !super.isEnabled()) {
             super.setEnabled(true);
+            finishedQueriesListener.enableListener(true);
             queryInsightsService.stop(); // Ensures a clean restart
             queryInsightsService.start();
         } else if (!anyFeatureEnabled && super.isEnabled()) {
             super.setEnabled(false);
+            finishedQueriesListener.enableListener(false);
```
dzane17: `anyFeatureEnabled` checks if Top N or query metrics are enabled. finishedQueriesListener should be independent of Top N

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -109,6 +110,7 @@ public QueryInsightsListener(
         this.queryInsightsService = queryInsightsService;
         this.queryShapeGenerator = new QueryShapeGenerator(clusterService);
         this.principalExtractor = threadPool != null ? new PrincipalExtractor(threadPool) : null;
+        this.finishedQueriesListener = new FinishedQueriesListener(queryInsightsService);
```
dzane17: Construct `FinishedQueriesListener` in QueryInsightsPlugin.createComponents

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,199 @@
...
+                    // Cluster is ready, start polling
+                    pollingTask = threadPool.scheduleWithFixedDelay(
+                        LiveQueriesCache.this::pollRunningTasks,
+                        new TimeValue(100, TimeUnit.MILLISECONDS),
```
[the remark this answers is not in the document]
KishoreKicha14: Updated

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            ListTasksRequest request = new ListTasksRequest().setActions(SEARCH_ACTION).setDetailed(true);
+
+            client.admin().cluster().listTasks(request, new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
+                    try {
+                        logger.debug("LiveQueriesCache polling found {} total tasks", response.getTasks().size());
+                        Map<String, TaskInfo> currentTasks = new HashMap<>();
```
LilyCaroline17: I think currentTasks is unused? It's being added to on line 101 but then it's never accessed.

src/main/java/org/opensearch/plugin/insights/rules/model/LiveQueryRecord.java
LilyCaroline17: Is `LiveQueryRecord` used anywhere?

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
KishoreKicha14: Removed

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -214,6 +215,15 @@ public void setGroupingFieldTypeEnabled(Boolean fieldTypeEnabled) {
         this.groupingFieldTypeEnabled = fieldTypeEnabled;
     }
 
+    /**
+     * Enable or disable finished queries tracking
+     * @param enabled boolean flag
+     */
+    public void setFinishedQueriesEnabled(boolean enabled) {
```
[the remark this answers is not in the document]
KishoreKicha14: removed the method

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -256,12 +258,20 @@ public void onRequestStart(SearchRequestContext searchRequestContext) {}
 
     @Override
     public void onRequestEnd(final SearchPhaseContext context, final SearchRequestContext searchRequestContext) {
-        constructSearchQueryRecord(context, searchRequestContext);
+        String sharedRecordId = java.util.UUID.randomUUID().toString();
+        constructSearchQueryRecord(context, searchRequestContext, sharedRecordId);
+        if (!skipSearchRequest(searchRequestContext)) {
+            captureFinishedQuery(context, searchRequestContext, sharedRecordId);
+        }
```
dzane17: Can you use the same TopN record for live queries? Is there a major difference between `SearchQueryRecord` and `LiveQueryRecord`? Right now it is doing a lot of work twice (for example extracting the username, role, wlm group, etc)

src/main/java/org/opensearch/plugin/insights/core/service/FinishedQueriesCache.java
```diff
@@ -0,0 +1,79 @@
...
+            lastAccessTime = currentTime;
+            return List.of();
+        }
+        lastAccessTime = currentTime;
+        synchronized (finishedQueries) {
+            while (!finishedQueries.isEmpty() && (lastAccessTime - finishedQueries.peekFirst().finishTime) > FINISHED_QUERY_RETENTION_MS) {
+                finishedQueries.removeFirst();
+            }
+            return finishedQueries.stream()
+                .sorted((a, b) -> Long.compare(b.finishTime, a.finishTime))
```
dzane17: I thought `finishedQueries` is already ordered by `finishTime`. Is this sort necessary? Use descending iterator if reverse order is needed.

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,214 @@
...
+                return;
+            }
+
+            ListTasksRequest request = new ListTasksRequest().setActions(SEARCH_ACTION).setDetailed(true);
+
+            client.admin().cluster().listTasks(request, new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
+                    try {
+                        logger.debug("LiveQueriesCache polling found {} total tasks", response.getTasks().size());
```
dzane17: Don't log every iteration

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,166 @@
...
+    private void startPolling() {
+        try {
+            // Check if cluster is ready before starting polling
+            client.admin().cluster().listTasks(new ListTasksRequest(), new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
+                    // Cluster is ready, start polling
+                    pollingTask = threadPool.scheduleWithFixedDelay(
+                        LiveQueriesCache.this::pollRunningTasks,
+                        new TimeValue(10, TimeUnit.MILLISECONDS),
```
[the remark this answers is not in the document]
KishoreKicha14: now it is 500ms do you think still it is less?

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,217 @@
...
+            }
+        } catch (Exception e) {
+            logger.debug("Failed to extract user info for task {}: {}", task.getTaskId(), e.getMessage());
+        }
+
+        return new SearchQueryRecord(task.getStartTime(), measurements, attributes, task.getTaskId().toString());
+    }
+
+    public List<SearchQueryRecord> getCurrentQueries() {
+        logger.debug("LiveQueriesCache returning {} cached queries", sortedQueries.length);
```
dzane17: No need to log

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            ListTasksRequest request = new ListTasksRequest().setActions(SEARCH_ACTION).setDetailed(true);
+
+            client.admin().cluster().listTasks(request, new ActionListener<ListTasksResponse>() {
+                @Override
+                public void onResponse(ListTasksResponse response) {
+                    try {
+                        Map<String, TaskInfo> currentTasks = new HashMap<>();
+                        TreeSet<SearchQueryRecord> topQueries = new TreeSet<>((a, b) -> {
```
dzane17: Rename var to `liveQueriesCache` or something else based on the next sentence.

I don't see this TreeSet ever get returned other than via `liveQueriesCacheSnapshot`.

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,214 @@
...
+                                topQueries.add(record);
+                                if (topQueries.size() > MAX_CACHE_SIZE) {
+                                    topQueries.pollLast();
+                                }
+                            }
+                        }
+
+                        logger.debug("LiveQueriesCache found {} search tasks, keeping top {}", searchTasks, topQueries.size());
+
+                        sortedQueries = topQueries.stream().limit(MAX_RETURNED_QUERIES).toArray(SearchQueryRecord[]::new);
```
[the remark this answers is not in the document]
dzane17: Rename `sortedQueries` to `liveQueriesCacheSnapshot`.

src/main/java/org/opensearch/plugin/insights/rules/transport/live_queries/TransportLiveQueriesAction.java
```diff
@@ -45,16 +47,44 @@ public class TransportLiveQueriesAction extends HandledTransportAction<LiveQueri
...
+        this.queryInsightsService = queryInsightsService;
     }
 
     @Override
     protected void doExecute(final Task task, final LiveQueriesRequest request, final ActionListener<LiveQueriesResponse> listener) {
+        // If use_live_cache=true, read directly from cache without ListTasks
+        if (request.isUseLiveCache()) {
+            try {
+                LiveQueriesCache liveCache = queryInsightsService.getLiveQueriesCache();
+                List<SearchQueryRecord> liveRecords = liveCache.getCurrentQueries();
```
dzane17: This returns the items in the cache array snapshot. What if the target taskId is not in this subset?

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Cache for live running queries
+ */
+public class LiveQueriesCache {
+
+    private static final Logger logger = LogManager.getLogger(LiveQueriesCache.class);
+    private static final int MAX_CACHE_SIZE = 1000;
+    private static final int MAX_RETURNED_QUERIES = 100;
+    private static final String SEARCH_ACTION = "indices:data/read/search";
```
ansjcy: This gives you the coordinator node search tasks, without any shard search tasks. I though we want to include the data node search tasks too in this PR?

ansjcy:
There is a architectural issue on this PR, let me try to explain understanding (say we have a cluster with 100 nodes)
1. cache is initialized on ALL 100 nodes
2. when this `client.admin().cluster().listTasks` is called in `LiveQueriesCache` on one node, it fans out to all other 99 nodes, getting coordinator node level search tasks for all nodes
3. that means on each poll (500 ms?), we are making 100*99 = 9900 calls
4. Also, all the content stored on different nodes are all duplicated.
5. When you call live queries with cache enabed, this request is not fan out to all nodes, so it only reads cache from local node - which is fine because based on 1 and 2, we are storing duplicated info on call nodes.

IMO this design option is very costy in terms of number of requests and storage.  we should just call list tasks on each node instead of `.cluster().listTasks`. and during the live queries get call, we should fan out at that time.

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
LilyCaroline17: wait it's still here?

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,217 @@
...
+        pollingTask = threadPool.scheduleWithFixedDelay(
+            this::getRunningTasks,
+            new TimeValue(POLLING_INTERVAL_MS, TimeUnit.MILLISECONDS),
+            ThreadPool.Names.GENERIC
+        );
+        // Do initial poll
+        getRunningTasks();
+    }
+
+    public void stop() {
```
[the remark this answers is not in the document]
dzane17: Nevermind, you are already setting the entire cache = null in `startIdleCheck`

src/main/java/org/opensearch/plugin/insights/core/service/LiveQueriesCache.java
```diff
@@ -0,0 +1,217 @@
...
+                        attributes.put(Attribute.USERNAME, userInfo.getUserName());
+                    }
+                    if (userInfo.getRoles() != null && !userInfo.getRoles().isEmpty()) {
+                        attributes.put(Attribute.USER_ROLES, userInfo.getRoles().toArray(new String[0]));
+                    }
+                }
+            }
+        } catch (Exception e) {
+            logger.debug("Failed to extract user info for task {}: {}", task.getTaskId(), e.getMessage());
+        }
```
LilyCaroline17: I think there's a problem here where PrincipalExtractor is called with onResponse which is async which runs on a different thread than the request so then the user info is wrong/missing

PR #508, opened 2025-12-23
Add username and user roles to top n queries

LilyCaroline17:
### Description
Add username and user roles to top n queries attributes to capture user information in query insights about who ran the queries. Corresponding tests were added as well.

### Issues Resolved

### Testing
1. Create a test user with specific roles
```
% curl -k -X PUT "localhost:9200/_plugins/_security/api/internalusers/testuser" \
-H 'Content-Type: application/json' \
-u admin:admin \
-d '{
  "password": "TestPass123!",
  "backend_roles": ["admin", "readall"]
}'
```
2. Send a search request as the test user
```
% curl -k -u testuser:TestPass123! -X GET "localhost:9200/_search" -H 'Content-Type: application/json' -d'
{
  "query": { "match_all": {} }
}'
```
3. Verify username and user_roles are captured in query insights 
```
% curl -k -u admin:admin -X GET "localhost:9200/_insights/top_queries?pretty"
```
Backend roles are mapped to user roles:
admin -> all_access
readall -> readall
and all users get own_index. 
```
{
  "top_queries" : [
    {
      "timestamp" : 1766522933488,
      "id" : "d0463734-cf13-4daf-aa39-3acd20783081",
      "wlm_group_id" : "DEFAULT_WORKLOAD_GROUP",
      "search_type" : "query_then_fetch",
      "source" : {
        "query" : {
          "match_all" : {
            "boost" : 1.0
          }
        }
      },
      "indices" : [ ],
      "total_shards" : 3,
      "phase_latency_map" : {
        "expand" : 0,
        "query" : 0,
        "fetch" : 0
      },
      "task_resource_usages" : [
        {
          "taskId" : 575,
          "parentTaskId" : -1,
          "nodeId" : "hCHQuMTaTaScUtgf3aLVlQ",
          "taskResourceUsage" : {
            "cpu_time_in_nanos" : 213000,
            "memory_in_bytes" : 23168
          }
        }
      ],
      "username" : "testuser",
      "node_id" : "hCHQuMTaTaScUtgf3aLVlQ",
      "group_by" : "NONE",
      "labels" : { },
      "user_roles" : [
        "own_index",
        "all_access",
        "readall"
      ],
      "measurements" : {
        "cpu" : {
          "number" : 213000,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "memory" : {
          "number" : 23168,
          "count" : 1,
          "aggregationType" : "NONE"
        },
        "latency" : {
          "number" : 1,
          "count" : 1,
          "aggregationType" : "NONE"
        }
      }
    }
  ]
}

```

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
+++ b/src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
@@ -95,7 +95,7 @@ public class QueryInsightsPlugin extends Plugin implements ActionPlugin, Telemet
             new QueryInsightsExporterFactory(client, clusterService),
             new QueryInsightsReaderFactory(client)
         );
-        return List.of(queryInsightsService, new QueryInsightsListener(clusterService, queryInsightsService, false));
+        return List.of(queryInsightsService, new QueryInsightsListener(clusterService, queryInsightsService, threadPool, false));
     }
 
     @Override
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -40,6 +40,7 @@ import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.core.index.Index;
 import org.opensearch.core.tasks.resourcetracker.TaskResourceInfo;
+import org.opensearch.plugin.insights.core.auth.PrincipalExtractor;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetric;
 import org.opensearch.plugin.insights.core.metrics.OperationalMetricsCounter;
 import org.opensearch.plugin.insights.core.service.QueryInsightsService;
@@ -51,6 +52,7 @@ import org.opensearch.plugin.insights.rules.model.SearchQueryRecord;
 import org.opensearch.plugin.insights.settings.QueryInsightsSettings;
 import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.tasks.Task;
+import org.opensearch.threadpool.ThreadPool;
 import reactor.util.annotation.NonNull;
 
 /**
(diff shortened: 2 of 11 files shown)


src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -93,11 +114,29 @@ public QueryInsightsListener(
         final ClusterService clusterService,
         final QueryInsightsService queryInsightsService,
         boolean initiallyEnabled
+    ) {
```
dzane17: I believe this constructor should only be used for tests since it is setting `threadpool = null`. Can we consider removing it to reduce the risk of someone using it for production in the future?

src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
```diff
@@ -62,6 +62,8 @@
  * Plugin class for Query Insights.
  */
 public class QueryInsightsPlugin extends Plugin implements ActionPlugin, TelemetryAwarePlugin {
+    private QueryInsightsListener queryInsightsListener;
```
ansjcy: why do we need to add a private field here?

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -292,6 +300,17 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
                     case WLM_GROUP_ID:
                         attributes.put(Attribute.WLM_GROUP_ID, parser.text());
                         break;
+                    case USERNAME:
+                        attributes.put(Attribute.USERNAME, parser.text());
```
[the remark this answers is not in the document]
LilyCaroline17: That's a good point. I think it'll keep things consistent and ensure that it's the correct types so maybe we should do this for all the cases in fromXContent in SearchQueryRecord?

src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
LilyCaroline17: Oh, I think this was left over from when I was trying to use extended plugin and the AttributeExtractor method. I'll remove it now that it's unnecessary.

src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
```diff
@@ -116,7 +126,12 @@ static void writeTo(final StreamOutput out, final Attribute attribute) throws IO
     @SuppressWarnings("unchecked")
     public static void writeValueTo(StreamOutput out, Object attributeValue) throws IOException {
         if (attributeValue instanceof List) {
-            out.writeList((List<? extends Writeable>) attributeValue);
+            List<?> list = (List<?>) attributeValue;
```
ansjcy: Do you think we need to check the out version here to avoid serialization issue? @dzane17 
This can potentially fail the top queries API during rolling upgrade.

LilyCaroline17: Wait, I realized I have USER_ROLES stored as a List<String>, but if I make it a String array (like indices since that was working already and is also an array of strings), I believe the existing serialization/deserialization should work, and then we don't have to make this update.

dzane17: I think it's okay since SearchQueryRecord attributes are in a Map. Looking at `Attribute.readAttributeMap()` and `SearchQueryRecord.writeTo()` methods, any additional attributes will be handled safely by readGenericValue and writeGenericValue.

dzane17: Yes, you are correct. I did not check the keys. I don't see a good solution with version checks since keys and values are written separately. Are we accepting the top queries API will return errors during upgrade then?

On a side note, we should add try catch inside of `readFromStream` to reduce the impact next time attributes are added.

PR #515, opened 2026-01-12
Track failed queries

### LINKED ISSUE #253: [FEATURE] Track Queries that error out in Query Insights
_A clear and concise description of what the problem is, e.g. I'm always frustrated when [...]._

We can consider tracking the following queries as part of Query Insights:
1. Queries that error out
2. Queries that do not complete due to timeout

Its hard for OpenSearch customers to figure out these queries in-case they want to modify them or retry in the future.

Additionally, With this approach when we integrate WLM and QI we can figure out tenants that are running erroneous or rogue queries.

For starters, we can probably enhance slow logs with an errorlogs feature to capture details about queries that error out and in the future think about how QI can help.
However, in QI this might entail keeping track of all the in-progress queries - not sure if we want to do this and if it will be scalable. We can also have a callback to QI on query failure if possible to make this easier.

imotov:
### Description
Adds support for keeping track of failed queries.

### Issues Resolved
Closes #253

I took a stab at adding some logic to be able to identify N most failing queries. I am not sure if this approach will work for you, but we need this feature, so I would love to iterate on it.


src/main/java/org/opensearch/plugin/insights/rules/model/MetricType.java
```diff
@@ -32,7 +33,11 @@ public enum MetricType implements Comparator<Number> {
     /**
      * JVM heap usage metric type
      */
-    MEMORY;
+    MEMORY,
+    /**
+     * Failure metric type
+     */
+    FAILURE;
```
ansjcy: what's the rationale to add failure as a Measurement Type instead of an Attribute? IMO it should be a boolean attribute to indicate the query "failed" or not.

src/main/java/org/opensearch/plugin/insights/QueryInsightsPlugin.java
```diff
@@ -145,6 +145,9 @@ public List<Setting<?>> getSettings() {
             QueryInsightsSettings.TOP_N_MEMORY_QUERIES_ENABLED,
             QueryInsightsSettings.TOP_N_MEMORY_QUERIES_SIZE,
             QueryInsightsSettings.TOP_N_MEMORY_QUERIES_WINDOW_SIZE,
+            QueryInsightsSettings.TOP_N_FAILURE_QUERIES_ENABLED,
```
ansjcy: I don't think settings are needed in this case, we can always enable it.

ansjcy (commented):
Thanks for the PR! I have a few questions related to the implementation.

src/main/java/org/opensearch/plugin/insights/rules/model/MetricType.java
imotov: My main motivation here was to be able to identify the top N failing queries. In OpenSearch, some failures are associated with high memory, CPU, and execution time, while others, such as incorrect Boolean query syntax, fail quickly and therefore score very low on these three metrics. My concern is that by relying on this tag, we might overlook the most frequently failing queries because they could be hidden under the noise of less frequent failures that consume more CPU, memory, and time. That said, I may not fully understand the capabilities of tags, so I’m open to converting this to a tag-based approach if this concern can be addressed.

dzane17: I recommend starting with option 1 (attribute tags) because it’s simpler and should cover most use cases.

I also don’t think “failed” belongs as a metric type alongside latency/CPU/memory. For option 2, I’d instead lean toward a dedicated failed-queries service within Query Insights. That service could be purpose-built (rather than following the Top N architecture) and could assume grouping is always enabled.

imotov: Sounds good. Let me explore option 1 then.

PR #525, opened 2026-01-22
Recent finshed queries

KishoreKicha14:
## Description

Implements a finished queries cache for the Live Queries API to capture recently completed queries with automatic idle timeout and cleanup.

### Core Implementation
- **FinishedQueriesCache**: New cache class that stores up to 1000 recently finished queries with 5-minute data retention
  - Captures query metrics (latency, CPU, memory), task IDs, user info, and cancellation status
  - Auto-stops after configurable idle timeout (default 5 minutes, range 2-10 minutes, or -1 to disable)
  - Lazy initialization on first API access
  - Returns up to 50 most recent queries

### API Changes
- **Live Queries API**: Added `use_finished_cache` parameter to include finished queries in results
  - Merges live queries from task manager with cached finished queries
  - Sorted by specified metric (latency/CPU/memory)

### Settings
- **`search.insights.live_queries.cache.idle_timeout`**: Dynamic cluster setting for cache idle timeout
  - Default: 5 minutes
  - Range: 2-10 minutes or -1 (disabled)
  - Cache auto-stops when idle beyond timeout

### Serialization
- Removed obsolete `cached` boolean from `LiveQueriesRequest` serialization
- Task ID stored in description field as `[task_id=<node_id>:<task_id>] <description>`

### Architecture
- Cache integrated into `QueryInsightsService` with lazy initialization
- Idle check task runs every minute to detect and stop expired cache
- `QueryInsightsListener` captures queries to cache without triggering lazy init
- Cache owns query capture logic (metrics calculation, user extraction)

## Testing
- All existing tests pass
- Cache auto-stop verified with idle timeout

## Example Usage

```bash
# Enable cache and set 2-minute idle timeout
PUT _cluster/settings
{
  "persistent": {
    "search.insights.live_queries.cache.idle_timeout": "2m"
  }
}

# Get live queries including finished queries from cache
GET /_insights/live_queries?use_finished_cache=true

### Issues Resolved

### CODE REVIEW

File: src/main/java/org/opensearch/plugin/insights/core/service/FinishedQueriesCache.java [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+    private static final int MAX_RETURNED_QUERIES = 50;
+
+    private final ArrayDeque<FinishedQuery> finishedQueries = new ArrayDeque<>();
+    private volatile long lastAccessTime = 0;
+    private volatile long retentionPeriodMs;
+    private volatile long idleTimeoutMs;
+    private final ClusterService clusterService;
+    private final PrincipalExtractor principalExtractor;
+
+    public FinishedQueriesCache(long retentionPeriodMs, ClusterService clusterService, ThreadPool threadPool) {
```
dzane17: Retention period does not need to be a param

src/main/java/org/opensearch/plugin/insights/rules/transport/live_queries/TransportLiveQueriesAction.java
```diff
@@ -125,14 +140,34 @@ public void onResponse(ListTasksResponse taskResponse) {
                         );
 
                         allFilteredRecords.add(record);
+
+                        // Add to finished queries cache
+                        FinishedQueriesCache cache = queryInsightsService.getFinishedQueriesCache();
+                        if (cache != null) {
+                            cache.addFinishedQuery(record);
```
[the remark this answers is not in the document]
ansjcy: +1, this is wrong. Why do we want to add unfinished tasks to finished cache? after a query is finished the same task will be added again to the cache in the listener, causing duplicated task records

src/main/java/org/opensearch/plugin/insights/rules/action/live_queries/LiveQueriesResponse.java
```diff
@@ -55,6 +74,10 @@ public List<SearchQueryRecord> getLiveQueries() {
     @Override
     public void writeTo(final StreamOutput out) throws IOException {
         out.writeList(liveQueries);
+        out.writeBoolean(useFinishedCache);
```
ansjcy: This will not be compatible with older versions and will cause serialization issues. Add version check here

PR #540, opened 2026-02-04
Tag failed queries

### LINKED ISSUE #253: [FEATURE] Track Queries that error out in Query Insights

imotov:
### Description
Adds support for tracking failed queries by setting the `failed` attribute. This PR supersedes #515.

### Issues Resolved
Closes #253

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
@@ -256,12 +256,12 @@ public final class QueryInsightsListener extends SearchRequestOperationsListener
 
     @Override
     public void onRequestEnd(final SearchPhaseContext context, final SearchRequestContext searchRequestContext) {
-        constructSearchQueryRecord(context, searchRequestContext);
+        constructSearchQueryRecord(context, searchRequestContext, false);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


imotov:
I’ll see if I can add success/failure counters to this.

imotov:
I started looking into @dzane17’s suggestion, and I think it should work. However, I agree with @ansjcy that it feels somewhat fragile since it relies on secondary attributes of the current data structures, which could easily break if those structures change in the future.

From a use‑case perspective, option 2 would be ideal for what I’m trying to achieve. Option 4 seems like a close approximation and should let me address the most pressing issues. Options 1 and 3 would still be useful, but they’re less suitable for my scenario because they don’t provide prioritization, especially since most of my queries change every time.

PR #541, opened 2026-02-04
Add RemoteRepositoryExporter to support exporting top n queries to remote repositories

LilyCaroline17:
### Description
Adds RemoteRepositoryExporter class to enable exporting top N query insights data to remote blob store repositories operating independently alongside existing local index and debug exporters. The implementation requires async multi-stream blob upload support, which is currently only available in OpenSearch's repository-s3 plugin. Support for additional repository types will be added as they implement async upload capabilities.

#### Settings: Added 3 cluster settings:

`search.insights.top_queries.exporter.remote.enabled` - Enable/disable remote export
`search.insights.top_queries.exporter.remote.repository `- Repository name
`search.insights.top_queries.exporter.remote.path` - Base path for organizing files

##### Example
```
PUT _cluster/settings
{
  "persistent" : {
    "search.insights.top_queries.exporter.remote.repository" : "my-s3-repository",
    "search.insights.top_queries.exporter.remote.path" : "query-insights"
    "search.insights.top_queries.exporter.remote.enabled" : true
  }
}
```
### File Structure

The exporter creates JSON files organized by timestamp: `{path}/top-queries/yyyy/MM/dd/HH/mm'UTC'/{node-id}-{metric-type}.json`

### Testing

Before configuring the repository, you must configure a blob store repository with OpenSearch. This requires the corresponding plugin to be installed as well. 

#### Example for S3 repository

After installing the `repository-s3` plugin, configure your s3 repository:

```
PUT _snapshot/my-s3-repository
{
  "type": "s3",
  "settings": {
    "bucket": "my-query-insights-bucket",
    "region": "us-east-1",
    "base_path": "opensearch-query-insights"
  }
}
```

#### Local testing with LocalStack :

A new `integTestRemoteRepository` task has been added to test the remote repository exporter functionality with S3. To run the remote repository export integration tests: 

```
# Start LocalStack
docker run -d --name localstack -p 4566:4566 localstack/localstack:latest

# Create S3 bucket
aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket --region us-east-1

# Run integration test
./gradlew integTest --tests "RemoteRepositoryExporterIT" -Ds3.enabled=true -Daccess_key=test -Dsecret_key=test -Dsession_token=test

# Verify export
aws --endpoint-url=http://localhost:4566 s3 ls s3://test-bucket/query-insights/ --recursive

# Cleanup
docker rm -f localstack
```

#### New GitHub Actions workflow

`remote_repository_exporter.yml` validates S3 file export. 

### Issues Resolved
Closes [#545]


src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,220 @@
...
+        if (repository == null) {
+            throw new RepositoryMissingException(repositoryName);
+        }
+
+        if (!(repository instanceof BlobStoreRepository)) {
+            throw new IllegalArgumentException("Repository " + repositoryName + " is not a blob store repository");
+        }
+
+        BlobStoreRepository blobStoreRepository = (BlobStoreRepository) repository;
+        BlobPath blobPath = BlobPath.cleanPath().add(basePath).add("top-queries");
```
dzane17: I don't think `top-queries` is needed in the path

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,220 @@
...
+        return id;
+    }
+
+    /**
+     * Export a list of SearchQueryRecord to remote repository
+     *
+     * @param records list of {@link SearchQueryRecord}
+     */
+    @Override
+    public void export(final List<SearchQueryRecord> records) {
```
dzane17: Right now every data node will simultaneously export its local top queries to the same object path and overwrite each other. I think suffixing each key with node-id will solve this.

```
query-insights/2026/02/05/16/10/utc/{node-id}.json
```

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,270 @@
...
+            @Override
+            public void onResponse(Void unused) {}
+
+            @Override
+            public void onFailure(Exception e) {
+                if (currentRetry < MAX_RETRIES) {
+                    long backoffMs = INITIAL_BACKOFF_MS * (1L << currentRetry);
+                    logger.warn("Upload failed, retrying in {}ms (attempt {}/{}): {}", backoffMs, currentRetry + 1, MAX_RETRIES, fileName);
+                    try {
+                        Thread.sleep(backoffMs);
```
ansjcy: This could be dangerous because `asyncContainer.asyncBlobUpload` uses a generic thread pool for uploading, which can potentially be used by other plugins. If we do sleep here it will definitely block the other threads. 

One option is you can completely remove the retry logic. Or if you want retry, you should delegate this sleep and retry logic to our exporter thread.

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,270 @@
...
+ * Supports S3, Azure Blob Storage, Google Cloud Storage, and other blob store types.
+ * Operates independently of other exporters.
+ */
+public class RemoteRepositoryExporter implements QueryInsightsExporter {
+    private static final Logger logger = LogManager.getLogger(RemoteRepositoryExporter.class);
+    private static final int MAX_RETRIES = 3;
+    private static final long INITIAL_BACKOFF_MS = 100;
+    private final Supplier<RepositoriesService> repositoriesServiceSupplier;
+    private final ClusterService clusterService;
+    private String repositoryName;
```
ansjcy: fields like `repositoryName` and `basePath` , `asycBlobContainer` should be volatile. Otherwise if one threads changes the value other threads will potentially still use the old values.

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,270 @@
...
+     * Build object key: yyyy/MM/dd/HH/mm'UTC'/{node-id}.json
+     */
+    private String buildObjectKey() {
+        ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
+        String nodeId = clusterService.localNode().getId();
+        return dateTimeFormatter.format(now) + "/" + nodeId + ".json";
+    }
+
+    @Override
+    public void close() {}
```
ansjcy: should we close the blobcontainer references in close function?

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,270 @@
...
+
+    private void uploadAsyncWithRetry(AsyncMultiStreamBlobContainer asyncContainer, List<SearchQueryRecord> records, int retryCount)
+        throws IOException {
+        StringBuilder jsonBuilder = new StringBuilder();
+        for (SearchQueryRecord record : records) {
+            jsonBuilder.append(record.toXContentForExport(XContentFactory.jsonBuilder(), ToXContent.EMPTY_PARAMS).toString()).append("\n");
+        }
+        byte[] data = jsonBuilder.toString().getBytes(StandardCharsets.UTF_8);
+
+        String fileName = buildObjectKey();
```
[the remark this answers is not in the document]
LilyCaroline17: I can add it to the serialization of the record part and only do it once.

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,270 @@
...
+     */
+    public String getBasePath() {
+        return basePath;
+    }
+
+    /**
+     * Set enabled state
+     */
+    public void setEnabled(boolean enabled) {
+        this.enabled = enabled;
```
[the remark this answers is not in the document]
LilyCaroline17: I realized that the warning could have false positives if we update all of the settings in one go because the other two could be empty when enabled is set to true, so I think not having a warning would be okay.

src/main/java/org/opensearch/plugin/insights/core/exporter/RemoteRepositoryExporter.java
```diff
@@ -0,0 +1,299 @@
...
+            throw new IllegalArgumentException("Repository " + repositoryName + " does not support async upload");
+        }
+
+        return (AsyncMultiStreamBlobContainer) blobContainer;
+    }
+
+    /**
+     * Build object key: yyyy/MM/dd/HH/mm'UTC'/{node-id}.json
+     */
+    private String buildObjectKey() {
```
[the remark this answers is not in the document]
LilyCaroline17: I'll make it so that each file will have the metric type appended to the end of the file name so that they don't overwrite each other!

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -183,6 +188,17 @@ public QueryInsightsService(
                 (localIndexLifecycleManager::setDeleteAfterAndDelete),
                 (localIndexLifecycleManager::validateDeleteAfter)
             );
+        queryInsightsExporterFactory.createRemoteRepositoryExporter(
```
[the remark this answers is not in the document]
LilyCaroline17: Should be fixed by not calling validation upon creation, but added try catches to update repository and path if users do that when a node is (re)starting up.

PR #549, opened 2026-02-13
Add recommendation data models for rule-based recommendation engine

ansjcy:
### Description
Add recommendation data models for rule-based recommendation engine.

New Data models:
  - Direction - enum for impact direction (INCREASE, DECREASE, NEUTRAL)
  - Impact - per-dimension impact with direction and optional numeric value for future magnitude estimation
  - RecommendationType - enum for recommendation categories (QUERY_REWRITE, INDEX_CONFIG, FEATURE_ENABLE, CORRECTNESS)
  - Action - recommended action with name, hint, documentation URL, and suggested query
  - ImpactVector - estimated impact across latency, CPU, memory, and correctness dimensions, using Impact objects for extensibility
  - Recommendation - top-level model combining rule metadata, action, impact, confidence, and metadata

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/recommendations/ImpactVector.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/recommendations/ImpactVector.java
@@ -9,7 +9,6 @@
 package org.opensearch.plugin.insights.rules.model.recommendations;
 
 import java.io.IOException;
-import java.util.Locale;
 import java.util.Objects;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -21,11 +20,10 @@ import org.opensearch.core.xcontent.XContentBuilder;
  * Represents the estimated impact of applying a recommendation
  */
 public class ImpactVector implements ToXContentObject, Writeable {
-    private final Direction latency;
-    private final Direction cpu;
-    private final Direction memory;
-    private final Direction correctness;
-    private final double confidence;
+    private final Impact latency;
+    private final Impact cpu;
+    private final Impact memory;
+    private final Impact correctness;
     private final String estimatedImprovement;
 
     private ImpactVector(Builder builder) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/plugin/insights/rules/model/recommendations/ImpactVector.java
```diff
@@ -0,0 +1,234 @@
...
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.core.common.io.stream.Writeable;
+import org.opensearch.core.xcontent.ToXContentObject;
+import org.opensearch.core.xcontent.XContentBuilder;
+
+/**
+ * Represents the estimated impact of applying a recommendation
+ */
+public class ImpactVector implements ToXContentObject, Writeable {
+    private final Direction latency;
```
dzane17: In the future if `value` is added will we change this to `Map<> latency` or add a new field `float latencyValue`?

```
  "impact_vector": {
    "latency": {
      "value": "float",               // Predicted impact on query latency (percentage or absolute value)
      "direction": "string"           // Direction of impact (e.g., "increase", "decrease")
    },
    ...
  }
```

src/main/java/org/opensearch/plugin/insights/rules/model/recommendations/Recommendation.java
```diff
@@ -0,0 +1,312 @@
...
+ */
+public class Recommendation implements ToXContentObject, Writeable {
+    private final String id;
+    private final String ruleId;
+    private final String title;
+    private final String description;
+    private final RecommendationType type;
+    private final Action action;
+    private final ImpactVector impact;
+    private final double confidence;
```
dzane17: What is the difference between Recommendation confidence and ImpactVector confidence? Is recommendation whether the rule applies and impact is estimated effect?

src/main/java/org/opensearch/plugin/insights/rules/model/recommendations/ImpactVector.java
ansjcy: Good call - I refactored this so each dimension is now an Impact object with direction and an optional value field. The JSON output is:
```
  "latency": { "direction": "decrease" }
```
In the future, when we need magnitude later, it becomes:
```
  "latency": { "direction": "decrease", "value": 0.5 }
```

PR #551, opened 2026-02-18
Add streaming dimension to query categorization metrics

harshavamsi:
### Description
This PR works on top of https://github.com/opensearch-project/OpenSearch/pull/20530. We now pass down a streaming flag for a request into the search request context from OpenSearch. We can then read that flag while categorizing a query and then tag that request with streaming. Right now streaming only works for aggregations so updated the aggregations tag

### Issues Resolved


harshavamsi:
@rishabhmaurya @bowenlan-amzn @ansjcy for review

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
```diff
@@ -391,6 +391,7 @@ private void constructSearchQueryRecord(
                 userPrincipalContext,
                 null
             );
+            record.setStreaming(searchRequestContext.isStreamingRequest());
```
ansjcy: Do you need this stream flag in top queries record? Or just metrics? 

If you need it in top queries, you need to do more than this and set / get, you need to change the serialization, copy constructor, equals, hascode etc.

src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
```diff
@@ -122,8 +123,10 @@ public void incrementCounter(QueryBuilder queryBuilder, int level, Map<MetricTyp
      * @param value value to increment
      * @param tags tags
      * @param measurements metrics measurements
+     * @param isStreaming whether the query is a streaming request
      */
-    public void incrementAggCounter(double value, Tags tags, Map<MetricType, Measurement> measurements) {
+    public void incrementAggCounter(double value, Tags tags, Map<MetricType, Measurement> measurements, boolean isStreaming) {
+        tags.addTag(IS_STREAMING_TAG, String.valueOf(isStreaming));
```
ansjcy: The `is_streaming` tag is only added to incrementAggCounter, but NOT to incrementCounter (query type counter) or incrementSortCounter. If a streaming request has a query clause and a sort clause but no aggregation, no streaming tag will be emitted for those metrics. 

Is this intentional?

src/main/java/org/opensearch/plugin/insights/core/listener/QueryInsightsListener.java
harshavamsi: For the time being we just need it in the metrics, we might want to analyze top queries later but that is not part of the plan right now

Issue #520, opened 2026-01-15
[FEATURE] User and Role Based Access Control for Query Insights Data


ansjcy:
### Is your feature request related to a problem?
As organizations adopt Query Insights in production multi-tenant environments, there is growing demand for more granular access control. Currently, the Top Queries API (`GET /_insights/top_queries`) is accessible to admin users or users with the `query_insights_all_access` security role, which provides visibility to all queries across the cluster.

For enterprise deployments with multiple teams sharing a cluster, organizations have requested the ability to scope query insights visibility based on user identity or team membership. This would enable:

- **Self-service debugging**: Development teams could independently access and analyze their own query performance without requiring admin privileges or support tickets
- **Multi-tenant isolation**: Teams working on different projects or serving different customers could maintain data privacy by viewing only queries relevant to their scope
- **Principle of least privilege**: Organizations could grant query insights access more broadly while maintaining appropriate security boundaries

This enhancement would make Query Insights more suitable for production environments where multiple teams or applications share the same OpenSearch cluster.

### What solution would you like?
Implement **user-based and role-based data filtering** for Query Insights APIs, allowing users to view only the queries they have permission to see. The solution provides multiple filtering modes through a single configuration setting.

#### 1. **Filter Modes**

A single setting `search.insights.top_queries.filter_by_mode` controls the filtering behavior with three possible values:

##### **Mode: `none` (Default - Current Behavior)**
- No filtering applied
- Admin users and users with `query_insights_all_access` role see all queries
- Maintains backward compatibility

##### **Mode: `username`**
- Users can only see queries they personally initiated
- Each query record is filtered by matching the `username` attribute
- For individual developers debugging their own queries, "Show me only my queries"

##### **Mode: `backend_roles`**
- Users can see queries from any user who shares at least one backend role
- Queries are filtered based on intersection of backend roles
- Ideal for team-based access where team members share backend roles
- Supports multi-team deployments where roles represent teams/departments
- Use case: "Show me queries from my team"

#### 2. **Cluster Settings**

```json
PUT _cluster/settings
{
  "persistent": {
    "search.insights.top_queries.filter_by_mode": "username"
  }
}
```

**Setting Details:**
- `search.insights.top_queries.filter_by_mode` (string, default: `"none"`)
  - Valid values: `"none"`, `"username"`, `"backend_roles"`
  - Dynamic setting - can be changed at runtime without restart
  - `"none"` maintains current behavior (backward compatible)

[...]

PR #552, opened 2026-02-19
Implement Access Control for Query Insights Data

ansjcy:
### Description
Add different methods for filter top queries by user / roles, following designs in https://github.com/opensearch-project/query-insights/issues/520

- new `search.insights.top_queries.filter_by_mode` cluster setting that controls access to top query records based on user identity. Supported modes: none (default, no filtering), username (users see only their own queries), and backend_roles (users see queries from users sharing at least one backend role).
- Users with the all_access role bypass filtering in all modes.
- Add backend_roles attribute to SearchQueryRecord by parsing it from the security user context string, alongside the existing username and user_roles fields.
- RBAC filtering is applied in `TransportTopQueriesAction` via a listener wrapper, covering both in-memory and historical (local index) query paths.
- Add runWithSecurity Gradle task for running OpenSearch with th

### Issues Resolved
related to https://github.com/opensearch-project/query-insights/issues/520

### Tests
See the integration test

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/exporter/LocalIndexExporter.java
@@ -21,6 +21,7 @@ import org.opensearch.ExceptionsHelper;
 import org.opensearch.ResourceAlreadyExistsException;
 import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.admin.indices.create.CreateIndexResponse;
+import org.opensearch.action.bulk.BulkItemResponse;
 import org.opensearch.action.bulk.BulkRequestBuilder;
 import org.opensearch.action.bulk.BulkResponse;
 import org.opensearch.action.index.IndexRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/plugin/insights/rules/transport/top_queries/TransportTopQueriesAction.java
```diff
@@ -202,7 +212,68 @@ private List<SearchQueryRecord> removeDuplicates(List<TopQueries> inMemoryTopQue
...
+            return delegate;
+        }
+        final FilterByMode mode = filterByMode;
+        final UserPrincipalInfo info = userInfo;
+        return new ActionListener<TopQueriesResponse>() {
+            @Override
+            public void onResponse(TopQueriesResponse response) {
+                try {
+                    List<TopQueries> filteredNodes = response.getNodes().stream().map(topQueries -> {
+                        List<SearchQueryRecord> filtered = TopQueriesRbacFilter.filterRecords(topQueries.getTopQueriesRecord(), mode, info);
```
dzane17: Right now RBAC filtering is done in-memory. There is the option to offload filtering of historical data to the reader search request since OpenSearch can efficiently handle this use-case.

dzane17 (approved):

Issue #532, opened 2026-01-30
[FEATURE] Rule-Based Query Recommendation Engine


ansjcy:
### Is your feature request related to a problem?
OpenSearch Query Insights currently identifies slow and resource-intensive queries but provides **no actionable guidance** on how to fix them. Users face several critical issues:

1. **Silent Failures**: Queries like `{"term": {"status": "Active"}}` on text fields return 0 results without warnings
2. **Performance Issues**: Leading wildcards (`*search`) cause 100-1000x slowdowns with no indication why
3. **Safety Risks**: Sorting on text fields can trigger OOM crashes that bring down entire clusters
4. **Knowledge Barrier**: Fixing these issues requires deep OpenSearch expertise that most users lack

For example:
```
User sees: Query latency = 5,200ms in Top N Queries dashboard
User thinks: "Why is this slow? What should I do?"
User outcome: Spends hours searching documentation, may not find solution
```

### What solution would you like?
A **rule-based recommendation engine** integrated into Query Insights that:

1. **Automatically analyzes queries** from Top N Queries and potentially profiler requests
2. **Detects anti-patterns** using predefined rules (Rules TBD)
3. **Generates actionable recommendations** with:
   - Clear problem description
   - Impact assessment (latency, memory, correctness)
   - Specific fix with code examples
   - Confidence scores
4. **Surfaces recommendations** through:
   - Top N Queries dashboard (inline badges and panels)
   - Query profiler page (on-demand analysis)
   - REST API endpoints (see below for supports)

```mermaid
sequenceDiagram
    participant User
    participant Dashboard
    participant API as REST API
    participant Service as RecommendationService
    participant Context as QueryContext
    participant Rules as Rule Engine
    participant Cache as Metadata Cache

    User->>Dashboard: View Top N Queries
    Dashboard->>API: GET /_insights/top_queries?recommendations=true
    API->>Service: analyzeTopQueries(records)

    loop For each query record
        Service->>Context: build(record)
        Context->>Cache: getFieldType(field)
        Cache-->>Context: "text"
        Context->>Cache: getFieldCardinality(field)
        Cache-->>Context: 10000000

        Service->>Rules: evaluate(context)
        Rules->>Rules: match all active rules
        Rules-->>Service: List<Recommendation>

        Service->>Service: attach recommendations to record
        Service->>Service: store.put(queryHash, recommendations)
    end

    Service-->>API: List<QueryRecord with recommendations>
[...]

PR #555, opened 2026-02-24
Rule-based Recommendation Service Implementation

ansjcy:
### Description
- Introduce the recommendation engine framework for Query Insights, enabling rule-based analysis of search queries with actionable recommendations
- Add RecommendationService to manage rule registration, on-demand evaluation, caching, and configurable filtering (confidence threshold, max count)
- Add RecommendationRule interface and QueryContext for rule implementations to inspect query structure, extract typed query builders via visitor pattern, and resolve field types from index mappings
- Extract SourceString.toSearchSourceBuilder() utility and add SearchQueryRecord.setSearchSourceBuilder() to support SSB reconstruction on transported/historical records
-  Register four new cluster settings: search.insights.recommendations.{enabled, min_confidence, max_count, enabled_rules}

### Issues Resolved
Related to https://github.com/opensearch-project/query-insights/issues/532

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,8 +23,8 @@ buildscript {
       opensearch_build += "-SNAPSHOT"
     }
 
-    def os_platform = "linux"
-    def artifact_type = "tar"
+    os_platform = "linux"
+    artifact_type = "tar"
     if (System.properties['os.name'].toLowerCase().contains('windows')) {
       os_platform = "windows"
       artifact_type = "zip"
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/recommendations/RecommendationService.java
```diff
@@ -0,0 +1,324 @@
...
+                return new ArrayList<>();
+            }
+
+            // Evaluate all rules
+            List<Recommendation> recommendations = evaluateRules(analyzableRecord);
+
+            // Cache result (evict oldest entries if needed)
+            if (cache.size() >= MAX_CACHE_SIZE) {
+                String firstKey = cache.keySet().iterator().next();
+                cache.remove(firstKey);
```
dzane17: I don't think this eviction is correct. HashMap does not keep track of insertion order. Also I think LRU is slightly better than FIFO. You can try OpenSearch's built-in cache `org.opensearch.common.cache Cache<String, List<Recommendation>>` for this.

src/main/java/org/opensearch/plugin/insights/core/service/recommendations/RecommendationService.java
```diff
@@ -0,0 +1,324 @@
...
+
+    /**
+     * Constructor
+     * @param clusterService the cluster service
+     * @param xContentRegistry the named XContent registry for reconstructing SearchSourceBuilder
+     */
+    public RecommendationService(ClusterService clusterService, NamedXContentRegistry xContentRegistry) {
+        this.clusterService = clusterService;
+        this.xContentRegistry = xContentRegistry;
+        this.rules = new ArrayList<>();
```
dzane17: Rules ArrayList is not thread safe. Concurrent update and iteration can cause an exception.

src/main/java/org/opensearch/plugin/insights/core/rules/RecommendationRule.java
```diff
@@ -0,0 +1,66 @@
...
+
+package org.opensearch.plugin.insights.core.rules;
+
+import org.opensearch.plugin.insights.core.service.recommendations.QueryContext;
+import org.opensearch.plugin.insights.rules.model.recommendations.Recommendation;
+
+/**
+ * Interface for recommendation rules that analyze queries and generate recommendations
+ */
+public interface RecommendationRule {
```
dzane17: Id, name, desc, priority, enabled are common fields for all rules so feels like `RecommendationRule` should be an abstract class rather than interface.

src/main/java/org/opensearch/plugin/insights/core/service/recommendations/RecommendationService.java
ansjcy: Good point, let me refactor this to the internal Cache

dzane17 (approved):

PR #569, opened 2026-03-11
Add recommendations parameter to Top Queries API

ansjcy:
### Description
Adds ?recommendations=true query parameter to the top_queries endpoint that enriches each query record with inline recommendations generated by the recommendation service.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesResponse.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesResponse.java
@@ -9,6 +9,7 @@
 package org.opensearch.plugin.insights.rules.action.top_queries;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


src/main/java/org/opensearch/plugin/insights/rules/action/top_queries/TopQueriesResponse.java
```diff
@@ -109,6 +112,12 @@ public String toString() {
      */
     private void toClusterLevelResult(final XContentBuilder builder, final Params params, final List<TopQueries> results)
         throws IOException {
+        // Merge pre-computed recommendations from all node responses
+        final Map<String, List<Recommendation>> mergedRecommendations = new HashMap<>();
+        for (TopQueries tq : results) {
+            mergedRecommendations.putAll(tq.getRecommendations());
```
KishoreKicha14: `putAll` here will quietly drop recommendations if two nodes happen to have the same record ID — last one wins. Probably won't happen because of dedup, but using `putIfAbsent` or merging lists would be safer

dzane17 (approved):
Approved, thanks

PR #573, opened 2026-03-25
Handle unrecognized SearchQueryRecord attributes during stream deserialization

dzane17:
### Description

During rolling upgrades, newer nodes may send `SearchQueryRecord` attributes that older nodes don't recognize (e.g. `SOURCE_TRUNCATED`, `FAILED`). Currently `Attribute.readFromStream` calls `Attribute.valueOf()` which throws `IllegalArgumentException` on unknown values, causing the entire node response to fail.

This change makes deserialization forward-compatible by:
- Catching `IllegalArgumentException` in `readFromStream` and returning `null` for unrecognized attribute names
- Skipping unknown attributes in `readAttributeMap` while still consuming their values from the stream to maintain correct stream position

Unknown attributes are written via `writeGenericValue`, so `readGenericValue` correctly advances the stream past them. Attributes with custom serialization (`SOURCE`, `TASK_RESOURCE_USAGES`) already exist in all versions and will never be unknown.

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
@@ -14,6 +14,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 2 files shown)


LilyCaroline17:
As long as we don't add any more custom attributes in the future, it looks good to me.

PR #580, opened 2026-03-29
Add search document count metric

shivamg2017:
### Description
Add search document count metric
Metric to fetch doc count returned by search query.

Relevant Opensearch PR to update SearchRequestContext : https://github.com/opensearch-project/OpenSearch/pull/21037

metric: ImmutableMetricData{resource=Resource{schemaUrl=null, attributes={service.name="OpenSearch"}}, instrumentationScopeInfo=InstrumentationScopeInfo{name=org.opensearch.telemetry, version=null, schemaUrl=null, attributes={}}, name=search.query.doc_count.histogram, description=Histogram for the number of documents returned per search query, unit=1, type=EXPONENTIAL_HISTOGRAM, data=ImmutableExponentialHistogramData{aggregationTemporality=CUMULATIVE, points=[ImmutableExponentialHistogramPointData{getStartEpochNanos=1774810013641740000, getEpochNanos=1774810493652552000, getAttributes={}, getScale=20, getSum=12.0, getCount=4, getZeroCount=0, hasMin=true, getMin=3.0, hasMax=true, getMax=3.0, getPositiveBuckets=DoubleExponentialHistogramBuckets{scale: 20, offset: 1661953, counts: {1661953=4} }, getNegativeBuckets=EmptyExponentialHistogramBuckets{scale=20, offset=0, bucketCounts=[], totalCount=0}, getExemplars=[]}]}}

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
@@ -51,6 +51,7 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
     private SearchSourceBuilder searchSourceBuilder;
     private final UserPrincipalContext userPrincipalContext; // Private field for user extraction
     private boolean streaming;
+    private int hitsLength;
 
     /**
      * Timestamp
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
dzane17, line 252: Please add `hitsLength` to the copy constructor. I know the value is only used for categorization histograms, but we should still keep this method up to date.

dzane17 (commented):
Can you add tests?
`QueryInsightsListenerTests.testOnRequestEnd` should mock `searchRequestContext.getHitsLength()` and assert the value on the captured record, similar to how `isStreamingRequest()` / `isStreaming()` is tested today.
`SearchQueryCategorizerTests` should verify that `recordSearchDocCount` is called with the expected value when categorize runs.

PR #586, opened 2026-04-01
Add index attribute in script query metric

shivamg2017:
Add index attribute as a tag to query type counters and histograms in the search query categorizer. This enables per-index granularity when tracking query type metrics.

### Changes

- Extract indices from SearchQueryRecord attributes and pass them through the categorizer pipeline (SearchQueryCategorizer → SearchQueryCategorizingVisitor → SearchQueryCounters)
- Tag counters and histograms with the index dimension when indices are available
- When a query targets multiple indices, metrics are emitted per-index
- Fall back to existing behavior (no index tag) when indices are empty or null

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
@@ -104,18 +104,30 @@ public final class SearchQueryCounters {
         this.queryHandlers = new HashMap<>();
     }
 
+    private static final String INDEX_TAG = "index";
+
     /**
      * Increment counter
      * @param queryBuilder query builder
      * @param level level of query builder, 0 being highest level
      * @param measurements metrics measurements
+     * @param indices indices targeted by the query
      */
-    public void incrementCounter(QueryBuilder queryBuilder, int level, Map<MetricType, Measurement> measurements) {
+    public void incrementCounter(QueryBuilder queryBuilder, int level, Map<MetricType, Measurement> measurements, String[] indices) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/categorizer/SearchQueryCounters.java
```diff
@@ -104,18 +104,30 @@ public SearchQueryCounters(MetricsRegistry metricsRegistry) {
...
+     * @param indices indices targeted by the query
      */
-    public void incrementCounter(QueryBuilder queryBuilder, int level, Map<MetricType, Measurement> measurements) {
+    public void incrementCounter(QueryBuilder queryBuilder, int level, Map<MetricType, Measurement> measurements, String[] indices) {
         String uniqueQueryCounterName = queryBuilder.getName();
 
         Counter counter = nameToQueryTypeCounters.computeIfAbsent(uniqueQueryCounterName, k -> createQueryCounter(k));
-        counter.add(1, Tags.create().addTag(LEVEL_TAG, level));
-        incrementAllHistograms(Tags.create().addTag(LEVEL_TAG, level).addTag(QUERY_TYPE_TAG, uniqueQueryCounterName), measurements);
+        if (indices != null && indices.length > 0) {
```
ansjcy, line 120: I have concerns on the potential cardinality explosion in metrics. Each unique index value creates a distinct tag combination - On a cluster with, say, 5,000 indices and 10 query types across 3 levels, that's potentially 5000 × 10 × 3 = 150,000 counter series plus the same number of histogram series. Not sure if this will be a good idea.

Issue #351, opened 2025-06-04
[FEATURE] Cache the cluster state request for query insights reader


ansjcy:
### Is your feature request related to a problem?
Each time we invoke the reader, we make a request to the cluster state api to fetch all available indices. There are some overhead with this approach.

### What solution would you like?
 Ideally we should cache the index names and update the cache whenever the cluster state is changed

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

dzane17:
I was wondering if we can send the reader search request to all indices in the time window with `allow_no_indices=true`& `ignore_unavailable=true` params instead of pre-filtering. This way we offload the cluster state lookup into the search request execution which is done efficiently in memory.

dzane17:
Hey @quangdutran, thanks for volunteering. The approach is not finalized yet. Once we get confirmation, feel free to start.

With the method I proposed
1. Query Insights does not need cluster state
2. The reader request will target a list of all `top_queries-*` within a range like `top_queries-2025.01.01-91845,top_queries-2025.01.02-91846,top_queries-2025.01.03-91847,...`
3. Time range to search is the intersection of the `delete_after` window and request `to`/`from` dates.

quangdutran:
Hi @dzane17 @ansjcy 

I've created a PR for this based on the discussion, could you guys check?

PR #614, opened 2026-05-26
[Feature] Skip calling cluster state for existing indices, but just generate the indices name purely from time inputs (#351)

quangdutran:
### Description
Stop calling cluster state. Generate the index-name list locally from the date range

### Issues Resolved
Implement #351 

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
+++ b/src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
@@ -42,7 +42,7 @@ public final class LocalIndexReader implements QueryInsightsReader {
     private DateTimeFormatter indexPattern;
     private final NamedXContentRegistry namedXContentRegistry;
     private final String id;
-    private int deleteAfterDays;
+    private volatile int deleteAfterDays;
 
     /**
      * Constructor of LocalIndexReader
@@ -90,6 +90,11 @@ public final class LocalIndexReader implements QueryInsightsReader {
     public LocalIndexReader setIndexPattern(DateTimeFormatter indexPattern) {
         this.indexPattern = indexPattern;
         return this;
+    }
+
+
+    public void setDeleteAfterDays(final int deleteAfterDays) {
+        this.deleteAfterDays = deleteAfterDays;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
```diff
@@ -475,7 +475,8 @@ public void setExporterAndReaderType(final SinkType sinkType) {
             queryInsightsReaderFactory.createLocalIndexReader(
                 TOP_QUERIES_READER_ID,
                 DEFAULT_TOP_N_QUERIES_INDEX_PATTERN,
-                namedXContentRegistry
+                namedXContentRegistry,
+                clusterService.getClusterSettings().get(TOP_N_EXPORTER_DELETE_AFTER)
```
dzane17, line 479: `delete_after_days` is a dynamic setting, but `LocalIndexReader` only picks up its value here at creation time. If it's later increased (e.g. 7 -> 30), the reader will still contain the original value.

src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
```diff
@@ -134,24 +139,22 @@ public void read(
...
-            }
+        final ZonedDateTime retentionTime = now.minusDays(this.deleteAfterDays);
+        final ZonedDateTime effectiveStart = start.isBefore(retentionTime) ? retentionTime : start;
 
-            @Override
-            public void onFailure(Exception e) {
-                listener.onFailure(e);
-            }
-        });
+        if (effectiveStart.isAfter(finalEnd)) {
```
dzane17, line 149: Can we add unit tests for the new clamp logic? E.g. results bounded to the retention window, this empty-result early return, and `delete_after_days = 0` (retentionTime becomes `now`, so any past range returns empty).

src/main/java/org/opensearch/plugin/insights/core/service/QueryInsightsService.java
quangdutran: value change listener has been added

src/main/java/org/opensearch/plugin/insights/core/reader/LocalIndexReader.java
quangdutran: Test cases have been added

dzane17 (approved):

dzane17:
Looks good @quangdutran

PR #647, opened 2026-07-19
[Feature] Support SQL/PPL query source tracking and aggregation

KishoreKicha14:
Read SQL/PPL metadata headers from SearchTask and surface them in query-insights top queries and live queries views.

Changes:
- Read x-query-source, x-original-query, x-query-execution-id, and x-query-phases headers from task labels in QueryInsightsListener
- Aggregate multiple DSL queries sharing the same execution ID into a single top-N record with summed metrics and sub-query breakdown
- Fix labels fromXContent parser to use actual field names instead of hard-coding X-Opaque-Id
- Add SUB_QUERIES attribute for storing per-DSL detail (id, source, indices, measurements) on aggregated records
- Collect all unique indices across sub-queries for the parent record

### Description

### Issues Resolved

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
@@ -408,8 +408,9 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
                         XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, parser.currentToken(), parser);
                         Map<String, Object> labels = new HashMap<>();
                         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
+                            String labelKey = parser.currentName();
                             parser.nextToken();
-                            labels.put(Task.X_OPAQUE_ID, parser.text());
+                            labels.put(labelKey, parser.text());
                         }
                         attributes.put(Attribute.LABELS, labels);
                         break;
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
@@ -290,6 +290,19 @@ public class SearchQueryRecord implements ToXContentObject, Writeable {
                         MetricType metric = MetricType.fromString(fieldName);
                         measurements.put(metric, Measurement.fromXContent(parser));
                         break;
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -291,6 +290,19 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
                         MetricType metric = MetricType.fromString(fieldName);
                         measurements.put(metric, Measurement.fromXContent(parser));
                         break;
+                    case MEASUREMENTS:
```
ansjcy, line 293: is this the measurements used inside each metrics? Why there could be a measurement object at this level in the json?

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -408,11 +420,76 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
...
                         Map<String, Object> labels = new HashMap<>();
                         while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
+                            String labelKey = parser.currentName();
                             parser.nextToken();
-                            labels.put(Task.X_OPAQUE_ID, parser.text());
+                            labels.put(labelKey, parser.text());
                         }
                         attributes.put(Attribute.LABELS, labels);
                         break;
+                    case "sql_phases":
```
ansjcy, line 429: using `sql_phases` is not the best practice here

src/main/java/org/opensearch/plugin/insights/rules/model/SearchQueryRecord.java
```diff
@@ -408,11 +420,76 @@ public static SearchQueryRecord fromXContent(XContentParser parser) throws IOExc
...
+                            while (parser.nextToken() != XContentParser.Token.END_OBJECT) {
+                                String metricKey = parser.currentName();
+                                parser.nextToken();
+                                phaseMetrics.put(metricKey, parser.longValue());
+                            }
+                            sqlPhases.put(phaseName, phaseMetrics);
+                        }
+                        attributes.put(Attribute.SQL_PHASES, sqlPhases);
+                        break;
+                    case "sub_queries":
```
ansjcy, line 446: same as above, avoid using string directly here.

PR #666, opened 2026-08-10
Handle unknown Attribute enum values gracefully during rolling upgrades

### LINKED ISSUE #510: [BUG] Add version check when serializing and deserializing Attribute enum
### What is the bug?

We are adding new attribute keys in below PRs
https://github.com/opensearch-project/query-insights/pull/355
https://github.com/opensearch-project/query-insights/pull/389
https://github.com/opensearch-project/query-insights/pull/508

During rolling upgrades, when GET /_insights/top_queries is called and old coordinator broadcasts to all nodes, new node will send TopQueriesResponse with new attributes like USER_ROLES. Then old Coordinator tries to deserialize this attribute with Attribute.valueOf() and it will fail. But I believe it will be caught in transport layer, not causing the node to crash.

We should add some vesion protection when serializing / deserializing Attribute enum

Related discussions here: https://github.com/opensearch-project/query-insights/pull/508#discussion_r2654689241

### How can one reproduce the bug?
_Steps to reproduce the behavior._

### What is the expected behavior?
_A clear and concise description of what you expected to happen._

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

_Add any other context about the problem._

pangwangshu:
## Problem
During rolling upgrades, when a newer OpenSearch node sends a `TopQueriesResponse` 
containing new attribute keys (e.g., `USER_ROLES`, `BACKEND_ROLES`) to an older 
coordinator whose `Attribute` enum predates those constants, the deserialization 
fails. The older node calls `Attribute.valueOf()` on the unknown attribute name, 
which throws an `IllegalArgumentException`, causing the entire response to be 
rejected and corrupting metrics collection.

## Solution
Wrap the `Attribute.valueOf()` call in a try-catch block within 
`readAttributeMap()`. When an unknown attribute is encountered:
1. Catch the `IllegalArgumentException`
2. Read and discard the value using `readGenericValue()` to maintain byte alignment
3. Continue processing the rest of the map

This approach works because all attributes (except those with special codecs like 
`TASK_RESOURCE_USAGES`, `SOURCE`, and `GROUP_BY`) use the self-describing generic 
value format that can be read and skipped safely without knowledge of the attribute.

## Key Insights
- **Byte alignment**: Since `readAttributeMap()` processes a fixed-size sequence of 
  key-value pairs, skipping any entry would corrupt the stream for all subsequent 
  entries without this fix.
- **Backwards compatible**: The fix doesn't change behavior for clusters with uniform 
  versions (no version skew).
- **Future-proof caveat**: The generic skip cannot safely consume `List` values (which 
  use `out.writeList()`). Currently, no attributes use `List` values, but future ones 
  with `List` values would need explicit handling in both `readAttributeMap()` and 
  `readAttributeValue()`.

Fixes #510

### DIFF
--- a/src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
+++ b/src/main/java/org/opensearch/plugin/insights/rules/model/Attribute.java
@@ -204,7 +204,37 @@ public enum Attribute {
         Map<Attribute, Object> map = new HashMap<>(size);
 
         for (int i = 0; i < size; i++) {
-            Attribute key = readFromStream(in);
+            final Attribute key;
+            try {
+                key = readFromStream(in);
+            } catch (IllegalArgumentException e) {
+                // The attribute name doesn't match any constant this node's Attribute enum
+                // knows about. This happens during a rolling upgrade: a node running a newer
+                // version can add new Attribute constants (e.g. USER_ROLES, BACKEND_ROLES) and
+                // send them to an older-version coordinator whose enum predates them, and
+                // Attribute.valueOf() throws IllegalArgumentException for the unknown name.
+                //
+                // We can't just let this exception propagate: readAttributeMap is reading a
+                // fixed-size sequence of key/value pairs, so any entry we fail to fully consume
+                // corrupts the byte alignment for every entry after it in this stream, not just
+                // this one. Every attribute defined today other than the handful with bespoke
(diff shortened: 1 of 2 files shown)


xiaoyuan0821:
I think this PR doesn't actually solve the problem. You only modified the read implementation, but the old version's
  read implementation remains unchanged. The new version still writes the new ENUM, while the old version still reads
  through the old implementation — so in essence, nothing has really changed.

